PR #60, opened 2026-03-08
Frontend refactor

leo-Zhizhu:
Refactor frontend: Integrate TanStack Query for optimal data fetching
    
    - Replace manual AbortControllers and useEffect polls with useQuery
    
    - Implement intelligent caching and background refetch logic for dashboard and market data
    
    - De-duplicate API requests and improve Largest Contentful Paint metric


leo-Zhizhu:
Before:  network before

After: network after


Before: performance before

After: performance after

Chen-zexi (approved):

PR #66, opened 2026-03-14
65 voice input in chat

leo-Zhizhu:
## Summary
Adds voice input (Web Speech API) to the chat input bar. Users can start/stop speech recognition via a mic button, and toggle between English and Chinese recognition languages. The feature auto-syncs with the app's locale setting unless the user has explicitly overridden it.

## Changes

- Added mic button + EN/CN language toggle to `chat-input.tsx` using Web Speech API (`SpeechRecognition` / `webkitSpeechRecognition`)
- Speech language preference persisted in `localStorage` with manual-override tracking
- Added `safeLocalStorage` utility in `lib/utils.ts` — catches localStorage errors (private browsing, storage quota exceeded)
- Refactored `threadStorage.ts` and `MarketView/prefs.ts` to use `safeLocalStorage` instead of inline try/catch
- Added `chat.voice.*` i18n keys to `en-US.json` and `zh-CN.json`
- Added `@types/dom-speech-recognition` for TypeScript types
- Added CSS pulse animation for active recording state

## Test Plan

- Test via web frontend


leo-Zhizhu:
@claude I've resolved the issues. Please check if the codes are ready to merge

Chen-zexi:
### Code review

Found 3 issues:

1. **Transcript duplication/corruption on longer dictation.** The `onresult` handler loops from `i = 0` on every event instead of starting at `event.resultIndex`. The Web Speech API's `event.results` is a cumulative list -- results below `event.resultIndex` are already finalized and should not be re-processed. With `continuous: true`, Chrome internally restarts recognition after silence gaps, resetting `event.results`. The loop-from-zero pattern will then drop or duplicate earlier speech. The fix is to iterate from `event.resultIndex` and accumulate finalized results in a separate ref.

https://github.com/ginlix-ai/LangAlpha/blob/30c8c44620ff0ccbde56a914065e5268dac2e294/web/src/components/ui/chat-input.tsx#L368-L376

2. **Ghost transcription after send.** `handleSend` clears the textarea (`setMessage('')`) but never calls `recognitionRef.current.stop()`. The mic button disappears (hidden by `!isLoading` at L1431), but the `SpeechRecognition` instance keeps running in the background. `onresult` continues firing, calling `setMessage(baseMessageRef.current + sessionTranscript)`, which fills the "empty" textarea with ghost text while the agent is streaming a response. When loading ends, the user sees unexpected text in their input. The fix is to call `recognitionRef.current?.stop()` and `setIsListening(false)` inside `handleSend`, or add a `useEffect` that stops recognition when `isLoading` becomes `true`.

https://github.com/ginlix-ai/LangAlpha/blob/30c8c44620ff0ccbde56a914065e5268dac2e294/web/src/components/ui/chat-input.tsx#L1430-L1432

3. **Orphaned recognition instance on rapid double-click.** `toggleListening` uses `isListening` from its `useCallback` closure (L340). Between clicking the mic button and React committing `isListening = true` (set asynchronously in `onstart`), a second click sees stale `isListening === false`, skips the stop branch, and creates a second `SpeechRecognition` instance. The first instance is orphaned -- `recognitionRef.current` is overwritten, so there is no handle to stop it. The orphaned instance keeps the microphone active with no UI indication, and its `onresult` callback continues calling `setMessage()`, corrupting input. The fix is to always stop any existing `recognitionRef.current` before creating a new instance, regardless of `isListening` state.

https://github.com/ginlix-ai/LangAlpha/blob/30c8c44620ff0ccbde56a914065e5268dac2e294/web/src/components/ui/chat-input.tsx#L339-L392

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Chen-zexi (approved):

PR #122, opened 2026-04-10
feat(llms): add Qwen3.6 Plus model support

Chen-zexi:
## Summary

- Add `qwen3.6-plus` model for DashScope (CN) with tiered pricing (≤256k and 256k-1M)
- Add `qwen3.6-plus-intl` model for DashScope International (Singapore) with tiered pricing
- Supports text, image, pdf, and video input modalities
- Pricing converted from CNY using the same rate as existing DashScope entries

## Test plan
- [x] All frontend tests pass (33 files, 360 tests)
- [x] JSON manifest files validated (models.json, providers.json)

### DIFF
--- a/src/llms/manifest/models.json
+++ b/src/llms/manifest/models.json
@@ -503,6 +503,20 @@
       "text"
     ]
   },
+  "qwen3.6-plus": {
+    "model_id": "qwen3.6-plus",
+    "provider": "dashscope",
+    "visible": true,
+    "input_modalities": [
+      "text",
+      "image",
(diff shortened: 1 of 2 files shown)


Chen-zexi:
The output ≤1M rate of 6.452 is correct. The DashScope International pricing page shows ¥44.965/1M tokens for output at the 256k-1M input tier, which converts to $6.452. The SG/CN ratio is not uniform across all tiers — the output ≤1M tier has a lower international markup than the other tiers. Reverted the pricing change.

PR #124, opened 2026-04-11
fix(widget): sanitize NaN/Infinity and catch runtime errors in inline widgets

Chen-zexi:
## Summary

Fixes two classes of widget rendering failures:

- **NaN/Infinity in JSON**: Python's `json.dumps()` outputs `NaN` for `float('nan')` (default `allow_nan=True`), common in pandas financial data (moving averages, etc.). Browser `JSON.parse()` rejects these. Fixed by patching `JSON.parse` inside the iframe and sanitizing `.json`/`.geojson`/`.topojson` data files on the backend.

- **Silent widget crashes**: Chart.js errors, missing libraries, and other runtime exceptions produced blank widgets with no feedback. Added an error overlay with a "Fix" button that sends the error back to the agent for self-correction.

**Defense-in-depth hardening from pre-landing review:**
- Fixed `-Infinity` regex (was producing `-null`, still invalid JSON)
- CSP meta tag moved before injected scripts
- Error overlay capped at 5, rendering guard against recursion
- Prompt truncation to 500 chars

## Pre-Landing Review
7 findings (2 critical, 5 informational) — all auto-fixed in second commit.

## Test plan
- [x] All backend tests pass (2380 passed)
- [x] All frontend tests pass (388 passed)
- [x] Linting clean (ruff + eslint)

### DIFF
--- a/src/ptc_agent/agent/tools/show_widget.py
+++ b/src/ptc_agent/agent/tools/show_widget.py
@@ -211,7 +211,7 @@ async def _resolve_data_files(
             _, ext = os.path.splitext(path)
             if ext.lower() in ('.json', '.geojson', '.topojson'):
                 value = re.sub(r'\bNaN\b', 'null', value)
-                value = re.sub(r'-?Infinity\b', 'null', value)
+                value = re.sub(r'(?<![A-Za-z_])-?Infinity\b', 'null', value)
         else:
             b64 = base64.b64encode(content).decode()  # type: ignore[arg-type]
             value = f"data:{mime};base64,{b64}"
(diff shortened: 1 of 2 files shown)


Chen-zexi:
Known tradeoff — `\bNaN\b` uses word boundaries so it only matches standalone "NaN", not substrings like "NaNometer". The remaining risk is "NaN" appearing as a standalone word inside a JSON string value (e.g. `"label": "NaN"`), which is vanishingly rare in pandas financial output. A full JSON token-stream parser would be the precise fix but adds complexity disproportionate to the risk. Accepting as-is.

PR #133, opened 2026-04-13
feat: add Lissajous Drift loading animation for chat streaming

Chen-zexi:
## Summary

Replace the MorphLoading (4-square morph) streaming indicator with a Lissajous Drift particle animation. The new component renders 68 SVG particles tracing a 3:4 Lissajous curve with pulsing amplitude, creating a fluid organic motion while the AI is generating text.

- **New component**: `web/src/components/ui/lissajous-loading.tsx` — self-contained SVG animation using requestAnimationFrame, proper cleanup on unmount, no dependencies
- **Swapped in**: `MessageList.tsx` streaming indicator now uses `LissajousLoading` instead of `MorphLoading`
- **Responsive**: `w-5 h-5` on mobile, `w-6 h-6` on desktop, with neutral grey color (`text-neutral-500` / `dark:text-neutral-400`)

Applies to both ChatAgent and MarketView (MarketPanel reuses MessageList).

## Test plan
- [x] All 419 Vitest tests pass (37 files)
- [x] ESLint clean (no new warnings)
- [x] Verified animation renders at correct size and alignment next to agent avatar
- [x] Verified responsive sizing on mobile viewport

### DIFF
--- a/web/src/components/ui/lissajous-loading.tsx
+++ b/web/src/components/ui/lissajous-loading.tsx
@@ -3,12 +3,6 @@ import { cn } from "@/lib/utils";
 
 const SVG_NS = "http://www.w3.org/2000/svg";
 
-const containerSizes: Record<string, string> = {
-  sm: "w-4 h-4",
-  md: "w-6 h-6",
-  lg: "w-8 h-8",
-};
-
 const PARTICLE_COUNT = 68;
 const TRAIL_SPAN = 0.34;
 const DURATION_MS = 6000;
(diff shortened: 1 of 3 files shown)


Chen-zexi:
## Performance tradeoff

MorphLoading uses CSS keyframe animations (compositor thread, zero JS cost). LissajousLoading uses requestAnimationFrame with imperative DOM updates on the main thread.

| | MorphLoading (old) | LissajousLoading (current) |
|---|---|---|
| Runtime JS | 0 | rAF loop |
| setAttribute/sec @60fps | 0 | 8,160 |
| Thread | Compositor | Main |

After optimization (static r/opacity, cx/cy only per frame), the cost is 136 setAttribute calls/frame — fine on any modern device, but not free like CSS keyframes. The Lissajous curve with particle trail can't be replicated in pure CSS since each particle position requires sin() computation at runtime.

Acceptable tradeoff for the visual quality. If streaming text fluidity regresses on low-end mobile, we can throttle to 30fps (halving cost to ~4k setAttribute/sec) or revert to MorphLoading.

PR #136, opened 2026-04-14
feat: auto-save Settings, remove save buttons

Chen-zexi:
## Summary

Remove manual save buttons from the Settings page. Every change now auto-saves immediately.

**Auto-save behavior:**
- **Name** (text input): debounced at 800ms
- **Timezone / Locale** (selects): saves immediately on change
- **Model preferences** (primary, flash, starred, advanced): debounced at 500ms

**UI changes:**
- Logout button moved to avatar row (right-aligned)
- Theme toggle restyled as card row, matching Voice Input style
- Save status indicators (Saving.../Saved/Error) replace save buttons

**New utility:**
- `useDebouncedSave` hook (`web/src/hooks/useDebouncedSave.ts`) wraps async save functions with configurable debounce, status tracking, and unmount cleanup

## Test Coverage

Tests: 42 → 43 (+1 new file, +6 tests)

```
useDebouncedSave.ts — 6/6 code paths tested (100%)
  ✓ idle status on init
  ✓ trigger() debounce + status transitions
  ✓ rapid trigger() coalescing
  ✓ flush() immediate fire + cancel pending
  ✓ error status + auto-reset
  ✓ cleanup on unmount
```

## Pre-Landing Review

3 issues found, all resolved:

| # | Severity | Issue | Action |
|---|----------|-------|--------|
| 1 | CRITICAL | Flush saves read stale ref (timezone/locale persisted old value) | Auto-fixed: eagerly update ref before flush |
| 2 | INFO | No timer cleanup on unmount in useDebouncedSave | Auto-fixed: added useEffect cleanup |
| 3 | INFO | `error` state never cleared after removing form submit handler | Fixed: added setError(null) in saveUserInfo |

PR Quality Score: 9/10

## Plan Completion

8/10 DONE, 1 CHANGED (API keys moved to /setup/method), 1 NOT DONE (toast errors — using inline status instead)

## Test plan

- [x] All Vitest tests pass (40 files, 445 tests)
- [x] TypeScript compiles with no errors
- [x] ESLint passes with 0 errors
- [x] Visual verification in browser: no save buttons, auto-save works, layout correct

### DIFF
--- a/web/src/hooks/__tests__/useDebouncedSave.test.ts
+++ b/web/src/hooks/__tests__/useDebouncedSave.test.ts
@@ -64,6 +64,26 @@ describe('useDebouncedSave', () => {
     expect(result.current.status).toBe('idle');
   });
 
+  it('queues a re-save when flush is called during an in-flight save', async () => {
+    let resolveFirst!: () => void;
+    const saveFn = vi.fn()
+      .mockImplementationOnce(() => new Promise<void>(r => { resolveFirst = r; }))
+      .mockResolvedValue(undefined);
+    const { result } = renderHook(() => useDebouncedSave(saveFn, 500));
+
+    // Start first save
+    act(() => { result.current.flush(); });
+    expect(saveFn).toHaveBeenCalledTimes(1);
+
+    // Second flush while first is in-flight — should queue, not start
+    act(() => { result.current.flush(); });
+    expect(saveFn).toHaveBeenCalledTimes(1);
+
+    // Complete first save — queued save should fire
+    await act(async () => { resolveFirst(); });
+    expect(saveFn).toHaveBeenCalledTimes(2);
+  });
+
   it('cleans up timers on unmount', async () => {
     const saveFn = vi.fn().mockResolvedValue(undefined);
     const { result, unmount } = renderHook(() => useDebouncedSave(saveFn, 500));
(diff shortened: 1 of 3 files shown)


Chen-zexi:
Addressed in fd0ef49:

**Concurrent execute() guard** — Added `runningRef` + `pendingRef`. If `flush()` fires while a save is in-flight, the second call queues and executes after the first completes. Test added for the queuing behavior (7 tests total now).

**`error` state block** — Intentional. The `{error && ...}` JSX block still serves avatar upload (`setError` at line 454) and preference reset (line 571) errors. User-info auto-save errors surface via the `userInfoSaveStatus === 'error'` inline indicator, which is the intended UX for the auto-save pattern.

**`flush` wrapper** — Intentional API design. `flush` communicates "save now" vs `execute` which is an internal implementation detail. Keeps the public API clean (`trigger`/`flush`/`status`) even though `flush` currently just delegates.

Chen-zexi:
Both addressed in ddd00c2:

1. **Unmount guard** — Added `mountedRef`. The cleanup effect sets it to `false`, and `execute()` checks it before `setStatus` calls and before re-executing queued saves. Prevents orphaned state updates and API calls after navigation.

2. **Name onBlur flush** — Added `onBlur={() => flushUserInfoSave()}` to the name input. Now if the user types a name and clicks away (tab, navigate, click another field), the pending 800ms debounce flushes immediately instead of being silently dropped by cleanup.

PR #145, opened 2026-04-18
fix: Opus 4.7 thinking-block schema + xhigh effort + tool-reasoning leak

Chen-zexi:
## Summary

Three related fixes for Opus 4.7 rollout.

**xhigh effort tier (`feat(reasoning)`)** — Introduces `xhigh` as a fourth reasoning effort alongside low/medium/high. Only the Anthropic adaptive path natively supports it; every other provider clamps `xhigh → high` via `_clamp()` in `reasoning.py`. Pydantic `ChatRequest.reasoning_effort` literal widens at the API boundary. Manifest adds `display: "summarized"` + `effort: "xhigh"` defaults to all three Opus 4.7 entries, plus `max_tokens` caps across the Anthropic family (Opus = 128k, Sonnet/Haiku = 64k). Frontend gates the X-High button behind a regex that accepts Opus 4.7/8/9/10+ and 5+ while rejecting 4.6 and earlier (forward-compat for future Opus releases without code changes).

**Sanitizer middleware (`fix(agent)`)** — Anthropic returns `messages.<i>.content.<j>.thinking.thinking: Field required` when an AI message carries a signature-only thinking block (no `thinking` text). This happens on Opus 4.7 when the stream produces a redacted/empty reasoning segment that LangGraph then persists. `AnthropicThinkingSanitizerMiddleware` runs innermost in both main and subagent chains, scans AI messages, and injects `thinking: ""` into orphan blocks while leaving everything else untouched by identity. Safe no-op for non-Anthropic providers (they don't emit thinking blocks).

**Tool-reasoning leak (`fix(streaming)`)** — Tools with internal LLMs (WebFetch, etc.) were emitting reasoning chunks under `agent: "tools:<uuid>"` with mis-ordered lifecycle signals, which the frontend painted as a second phantom Reasoning panel on the main agent's turn. `_process_message_chunk` now gates reasoning emission when `agent_name == "tools"` or starts with `"tools:"`. Subagents (`task:*`, `research:*`) are unaffected.

## Test Coverage

- **Backend (2515 passed)** — `test_reasoning.py` parametrized for xhigh + 5 clamping tests per provider, `test_chat_models.py` widens reasoning_effort loop, `test_streaming_handler.py` adds `TestToolAgentReasoningSuppression` (4 cases), `test_anthropic_thinking_sanitizer.py` new (18 cases: block-level repair, message scan, list identity, async + sync override contract).
- **Frontend (467 passed, +21 new)** — `modelCapabilities.test.ts` covers the xhigh regex against 11 accept + 9 reject + 3 null-ish cases to lock the forward-compat pattern.

## Pre-Landing Review

No P0/P1 findings. Three concerns verified:
- Sanitizer only mutates `{"type":"thinking"}` blocks; identity-preserving elsewhere.
- `_clamp()` leaves xhigh raw only on the adaptive Anthropic path.
- Tool-agent gate uses an exact `"tools"` match + `"tools:"` prefix, never touches `task:*` / `research:*`.

## Test plan

- [x] Backend unit tests: 2515 passed
- [x] Frontend vitest: 467 passed
- [x] TypeScript check: clean
- [ ] Manual smoke against Opus 4.7 with `reasoning_effort: xhigh` in prod traffic

### DIFF
--- a/src/llms/manifest/models.json
+++ b/src/llms/manifest/models.json
@@ -60,11 +60,13 @@
       "pdf"
     ],
     "parameters": {
+      "max_tokens": 128000,
       "thinking": {
-        "type": "adaptive"
+        "type": "adaptive",
+        "display": "summarized"
       },
       "output_config": {
-        "effort": "medium"
+        "effort": "xhigh"
       }
     }
   },
@@ -78,6 +80,7 @@
       "pdf"
     ],
     "parameters": {
+      "max_tokens": 64000,
       "thinking": {
         "type": "adaptive"
       }
@@ -93,6 +96,7 @@
       "pdf"
     ],
     "parameters": {
+      "max_tokens": 64000,
       "thinking": {
         "type": "enabled"
       }
@@ -759,6 +763,7 @@
       "pdf"
     ],
     "parameters": {
+      "max_tokens": 64000,
       "thinking": {
         "type": "adaptive"
       }
@@ -774,11 +779,13 @@
       "pdf"
     ],
     "parameters": {
+      "max_tokens": 128000,
       "thinking": {
-        "type": "adaptive"
+        "type": "adaptive",
+        "display": "summarized"
       },
       "output_config": {
-        "effort": "medium"
+        "effort": "xhigh"
       }
     }
   },
@@ -792,6 +799,7 @@
       "pdf"
     ],
     "parameters": {
+      "max_tokens": 64000,
       "thinking": {
         "type": "enabled",
         "budget_tokens": 8000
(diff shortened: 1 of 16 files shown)


Chen-zexi:
Acking both non-blocking items — keeping as-is:

**1. Regex `[5-9]` on major:** intentional. Anthropic's version progression (3 → 3.5 → 4 → 4.5 → 4.6 → 4.7) means a two-digit major is decades out, and the naming pattern bumps the minor long before hitting that boundary. The `\d{2,}` on the 4.x minor axis is the realistic forward-compat lane; widening the major to `\d{2,}` expands the acceptance set beyond meaningful names.

**2. Orphaned `reasoning_active` state on tool-agent paths:** no bug today. Both entry points (`status_info` at L849 and `reasoning_content_from_kwargs` at L872) are symmetrically guarded on `if not is_tool_agent:`, so tool agents can't enter `self.reasoning_active` in the first place. If a future refactor adds a third entry point without the guard, that's where the bug would land — not in the cleanup path. Prefer to keep the set-add invariant tight rather than scatter defensive discards.

PR #151, opened 2026-04-20
feat(ux): signal archived sandbox restore + extend workspace lifecycle

Chen-zexi:
## Summary

Cold-restore from archived Daytona sandbox takes ~60–90s (snapshot download + container reboot). Until now the UI showed the same "Starting workspace…" copy as a warm ~2s restart, so users thought the app was hung. This PR plumbs a callback from `PTCSandbox.reconnect` back up to the chat SSE generator so the frontend can show honest copy the moment we know the sandbox is archived.

**Backend**
- `on_state_observed` callback threads through `Session.initialize` / `initialize_lazy` → `PTCSandbox.reconnect` → fires once with the state string (`"archived"`, `"running"`, ...) right after `get_state`
- Chat SSE generator races the callback against session init with `asyncio.Event` + `wait_for(5.0)`. If state is `archived`, emits a second `workspace_status` event with `sandbox_state: "archived"` between the initial `starting` and the final `ready`. No FIRST_COMPLETED race loop (avoids the `asyncio-shield-cancel-bypass` precedent)
- Orphan-task guard: client disconnect during the refinement window cancels the in-flight `session_task` instead of leaking it

**Config**
- `auto_archive_interval`: 24h → 7d (keep workspaces in fast-restart `stopped` state longer; stopped sandboxes cost \$0/hr)
- `auto_delete_interval`: 7d → 90d (extend total dormant lifetime)
- `DaytonaConfig` dataclass is now the Python SSoT for defaults; `agent.py` reads from it instead of duplicating literals

**Frontend**
- `workspaceStarting` state widened to `false | 'starting' | 'archived'`
- Inline copy switches to "Restoring workspace from storage, this usually takes about 60 seconds…" on the archived path
- Hover card (`@radix-ui/react-hover-card`) explaining why it's slow + what's preserved, in en-US + zh-CN

## Test Coverage

- New: `tests/unit/server/handlers/test_ptc_workflow_workspace_status.py` — 6 tests pinning the SSE contract (archived 3-event, stopped/running 2-event, recovery timeout, warm 0-event, refinement ordering)
- New: `tests/unit/server/services/test_workspace_manager.py::TestOnStateObservedForwarding` — 4 assertions locking the callback-forwarding kwarg at every init call site (running, stopped+lazy, forced non-lazy, warm cache)
[...]

PR #157, opened 2026-04-21
fix(sandbox): stop error-state cascade from archived-restore failures

Chen-zexi:
## Summary

Production incident 2026-04-20 14:10-14:11: a failed Daytona archived-restore cascaded into ~75s of degraded behavior — 50s chat stall, workflow run on a broken session, 13s of 503s on `/files`, three "Future exception was never retrieved" asyncio warnings. One control-plane blip, an order of magnitude too much user-visible damage.

Four coordinated fixes + one opportunistic performance win + one logger regression.

**Workspace lifecycle**
- New `"starting"` status between `stopped` and `running`. Lazy-init flips to `starting`; Phase 2 promotes to `running` only after `ensure_sandbox_ready` actually succeeds. Consumers (`workspace_files.py`, `public.py`) route `starting` to the DB fallback so concurrent `/files` calls during restore no longer 503.
- Migration 009 adds `starting` to the `workspaces_status_check` CHECK constraint.
- Phase 2 error handling narrowed: `SandboxTransientError` + `has_failed()` → clear and re-raise. Post-init transient (sandbox already healthy) → best-effort retry. Plain `Exception` → legacy log-and-retry (regression-guarded).
- New `_clear_session` helper awaits `SessionManager.cleanup_session` so MCP and aiohttp provider close proactively, not on GC. Migrated all 5 broken-session clearing sites.
- Identity-guarded clear on both `SandboxGoneError` and `SandboxTransientError` paths (Phase 2 runs outside the workspace lock by design). `cleanup_session` itself is now identity-guarded on its final pop, fixing a latent `KeyError` under concurrent callers.
- `POST /start` returns 200 `"already starting"` for `starting`-state workspaces. `/files` write 409s include the actual status.

**Sandbox**
- Removed dead `asyncio.Future` from `_ensure_sandbox_connected` — the surrounding `_reconnect_lock` already fully serialized callers, so the Future only produced "Future exception was never retrieved" noise on every failed restore.
- Added `PTCSandbox.acreate_directories(paths)` — single `shlex`-quoted `mkdir -p` exec over all paths.

**File restore perf**
- Rewrote `FilePersistenceService.restore_to_sandbox` to bulk-mkdir unique parent dirs once, then upload through an `asyncio.Semaphore(16)` worker pool. On the incident workspace (230 files) this should cut `file_restore` from ~16s to ~2.5-3s. Per-file failures isolated via try/except tuples; fallback to parallel per-dir `acreate_directory` if bulk exec fails (ARG_MAX on huge workspaces).

**Unrelated regression fix**
- `src/server/handlers/chat/ptc_workflow.py:398` was passing structlog-style kwargs to a stdlib logger since PR #151. The fix had landed in #155 but was lost in a squash. Routed through `extra={}`.

## Test Coverage

- 14 new unit tests across sandbox, workspace, session, persistence, models.
- `tests/unit/server/app/test_workspace_files_routing.py` (new) pins plan item 8: concurrent failing lazy init + `/files` call → DB fallback, not 503.
- `tests/unit/server/services/test_file_persistence_restore.py` (new) pins the bulk-mkdir + semaphore-cap + per-file-failure behavior.
- No Future leak on reconnect failure; lock still serializes two concurrent `_ensure_sandbox_connected` callers.
- All 2744 unit tests pass on merged state.

## Pre-Landing Review

Cleared. Two rounds — initial review auto-fixed 4 UX strings and caught the P0 migration gap + P1 concurrent-clear race; second round caught stale `WorkspaceStatus` enum, bare 400 on `starting` state for `/start`, and `cleanup_session` KeyError under concurrent callers. All resolved in this PR.

## Test plan

- [x] `uv run pytest tests/unit/` → 2744 passed
- [x] `uv run ruff check` → clean on touched files
- [ ] Cold-restart an archived workspace and watch `file_restore` in `[SESSION_TIMING]` drop from ~16s to ~3s
- [ ] Force a Daytona control-plane conflict during restore; confirm `/files` returns `source: "database"` instead of 503 storm; confirm no zombie workflow runs on the failed session; confirm DB status stays at `starting` and next request recovers cleanly
- [ ] Grep logs for `"Future exception was never retrieved"` — should be zero after deploy

## Deploy ordering

Run migration 009 before rolling out the code (code writes `status='starting'`; CHECK constraint must accept it first). Standard `make migrate` → deploy ordering.

### DIFF
--- a/migrations/versions/009_add_starting_workspace_status.py
+++ b/migrations/versions/009_add_starting_workspace_status.py
@@ -1,0 +1,51 @@
+"""Add 'starting' to workspaces.status CHECK constraint.
+
+Used by the lazy-restart path in WorkspaceManager: stopped -> starting -> running.
+The intermediate 'starting' lets /files and /public callers fall back to the DB
+while Phase 2 (ensure_sandbox_ready + asset sync) is still resolving.
+
+Revision ID: 009
+"""
+
+from alembic import op
+
+
+revision = "009"
+down_revision = "008"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute("""
+        ALTER TABLE workspaces
+            DROP CONSTRAINT IF EXISTS workspaces_status_check
+    """)
+    op.execute("""
+        ALTER TABLE workspaces
+            ADD CONSTRAINT workspaces_status_check
+            CHECK (status IN (
+                'creating','running','starting','stopping',
+                'stopped','error','deleted','flash'
+            ))
+    """)
+
+
+def downgrade() -> None:
+    # Any rows parked in 'starting' would violate the older constraint; coerce
+    # them to 'stopped' so the next request re-enters the restart flow cleanly.
+    op.execute("""
+        UPDATE workspaces SET status = 'stopped' WHERE status = 'starting'
+    """)
+    op.execute("""
+        ALTER TABLE workspaces
+            DROP CONSTRAINT IF EXISTS workspaces_status_check
+    """)
+    op.execute("""
+        ALTER TABLE workspaces
+            ADD CONSTRAINT workspaces_status_check
+            CHECK (status IN (
+                'creating','running','stopping',
+                'stopped','error','deleted','flash'
+            ))
+    """)
--- a/src/ptc_agent/core/sandbox/ptc_sandbox.py
+++ b/src/ptc_agent/core/sandbox/ptc_sandbox.py
@@ -7,7 +7,7 @@ import json
 import shlex
 import textwrap
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from dataclasses import dataclass, field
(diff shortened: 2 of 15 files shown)


Chen-zexi:
Both verified as real. Both fixed in `a04a34d3`.

**Finding 1** — `session.py`: `"Session removed"` log is now inside the `if` branch; the identity-guarded skip logs `"Session cleaned up (replacement retained)"` instead. No more misleading log when a concurrent request wins the race.

**Finding 2** — `workspace_manager.py::_clear_session`: added an optional `evict_session` keyword arg that identity-guards the local `self._sessions.pop`. Threaded it through the two Phase-2 call sites (`SandboxGoneError` and `SandboxTransientError` outside the workspace lock). Inside-lock call sites omit it and keep the unconditional pop — the lock already prevents the race there.

Tests: 2744 pass. No new regressions.

PR #159, opened 2026-04-21
feat(tools): enforce TodoWrite schema at the tool signature via pydantic

### LINKED ISSUE #160: Bug: SSE error-path in todo sse_middleware sends unnormalized todos to frontend
## Summary

In `src/ptc_agent/agent/middleware/todo_operations/sse_middleware.py` (line 155), the `except` block that emits a `"failed"` SSE event sends the raw, unnormalized `todos` list directly:

```python
"payload": {
    "todos": todos,      # raw — not whitelisted, status not lowercased
    "total": len(todos),
    "error": str(e),
},
```

PR #159 normalized the success path so that only `{content, activeForm, status}` fields are emitted and status is lowercased. The error path was not updated and still forwards raw LLM-supplied dicts, which may include:
- Legacy fields (`id`, `created_at`, `updated_at`) the frontend does not expect
- Uppercase status values (`"COMPLETED"`, `"PENDING"`) that break the frontend's strict `status === 'pending'` equality checks
- A `total` count that may diverge from `completed + in_progress + pending` if any non-dict items are present

## Fix

Apply the same whitelist + lowercase normalization to the failed-event `todos` payload, or extract the normalization step before the `try` block so both paths share the same `normalized_todos`.

This was found in unchanged code during review of PR #159.

Chen-zexi:
## Summary

Tightens the `TodoWrite` tool so the LLM hits Pydantic at the tool boundary instead of getting a silent "validation warnings appended to a success string" response. **Supersedes #156** — the upstream schema fix here makes most of #156 redundant, but the valuable defensive bits (middleware guard, middleware tests, frontend tautology cleanup) have been ported in (commits `984a8da7` + `9c41d3ec`). **Closes #160** (error-path normalization, commit `bbce68c9`).

**Schema enforcement**
- Tool signature changed from `List[Dict[str, Any]]` to `List[TodoItem]` (Pydantic v2 model with `content`/`activeForm`/`status` only, `min_length=1` on strings, `TodoStatus` enum).
- Invalid shapes now raise `ValidationError`, which LangChain surfaces to the LLM as an error `ToolMessage` — it can self-correct on the next turn.
- Removed the post-hoc `validate_todo_list_dict` call from the tool body. Pydantic at the boundary is authoritative.

**Schema hygiene**
- Stripped internal metadata (`id`, `created_at`, `updated_at`) from the agent-facing `TodoItem` — they leaked into the LLM schema but nothing in the codebase read them through the model.
- Dropped dead `to_dict`/`from_dict` methods.
- Explicit `model_config = ConfigDict(extra="ignore")` so a future base-class flip to `extra="forbid"` can't silently break in-flight LLM sessions.

**Prompt consolidation**
- Moved workflow rules from the tool's 85-line docstring into `task_workflow.md.j2`. Tool docstring is ~10 lines now.

**Middleware hardening** (partly ported from #156, error-path fix closes #160)
- Payload whitelisted to `{content, activeForm, status}` before SSE emission — strips LLM-sent legacy fields from the wire + persistence layer, lowercases status so the frontend's `status === 'pending'` strict equality holds regardless of case.
- Top-level `isinstance(list)` guard + `logger.warning` — protects the failed-event branch's `len(todos)` call from crashing on None/int payloads. Without this, a secondary TypeError would swallow the original ValidationError.
- **Error-path normalization** (#160): lifted the whitelist + lowercase + status-count loop above the `try` block so the failed event ships the same normalized shape as the success event. Previously the error path sent raw LLM todos with legacy fields and uppercase status, breaking the frontend's strict equality checks on failure.

**Frontend cleanup** (ported from #156)
- Three tautological `todos || []` fallbacks removed from `streamEventHandlers.ts` and `historyEventHandlers.ts`.

## Relationship to #156

PR #156 was the narrow "guard the middleware against non-list payloads" follow-up to #153. This PR fixes the root cause one layer higher (the tool schema itself). What was kept from #156:

| #156 item | Status in #159 | Why |
|---|---|---|
| `isinstance(list)` guard in middleware | **Kept** (ported) | Still protects the failed-event branch's `len(todos)` call — pydantic's ValidationError path would otherwise crash on None/int. |
| Middleware test file (8 tests) | **Kept + extended to 13** | Middleware had zero coverage; tests close that gap and cover new whitelist + status normalization behavior (including error-path normalization from #160). |
| Frontend `todos \|\| []` cleanup | **Kept** | Tautological after #153, readability win. |
| Single-point-of-defense "isinstance at the top" as the primary fix | **Superseded** | Primary defense is now the tool schema (pydantic rejects bad shapes before middleware runs). The guard is a belt-and-braces layer. |

#156 can be closed. All three valuable pieces live here.

## Relationship to #160

#160 flagged that the SSE failed-event branch still shipped raw LLM input (legacy fields, uppercase status, non-dict-inflated `total`) after the success path was normalized. Fixed by lifting the normalization loop above the `try` block so both paths share the same `normalized_todos` / `status_counts`. Closes on merge.

## Test Coverage

Tests: 18 → 40 (+9 tool-level in /review, +13 middleware).
Full unit suite: **2784/2784 pass**. Ruff clean.

Tool-level coverage: all 4 return-message branches, all 3 schema-field rejections, case-insensitive status normalization, legacy-field tolerance, 5 invalid payload shapes, 5 non-dict list-element variants, error messages that name valid options.

Middleware coverage (was zero before): list pass-through with counts, status case normalization in payload (not just counter), legacy-field stripping from payload, non-list payload coercion (5 variants), non-TodoWrite pass-through, warning log capture, handler-exception branch without secondary crash, **failed-event payload whitelist + lowercasing**, **failed-event `total` drops non-dict items**.

## Pre-Landing Review

Specialists (Testing + Maintainability) ran during `/review`. 4 findings auto-fixed (backward-compat test, case-insensitive status test, non-dict list parametrize, bad-status assertion strengthening). 2 out-of-scope items flagged for follow-up: `manager.py:158` still writes stripped fields (helper has zero callers, pre-existing dead code), and a minor duplicated string in `tool.py`.

PR Quality Score: 9.0/10.

## Adversarial Review

Claude adversarial subagent + Codex adversarial challenge both ran. Converged on the same high-priority issue: the initial middleware lowercase fix was a half-measure (counter normalized, payload still shipped raw casing to the frontend). Fixed during this PR with payload-field whitelisting. Post-merge bot review caught the same issue on the error path (#160), also now fixed.

**Fixed during this PR:**
- Payload status normalization (not just counter).
- Explicit `ConfigDict(extra='ignore')`.
- Test docstring corrected to clarify `ToolArgumentParsingMiddleware` JSON-decodes stringified args in production.
- Top-level isinstance guard (from #156) protects failed-event branch.
- Error-path payload normalization (#160): failed events now whitelist + lowercase.

**Known and accepted:**
- Backward-compat for in-flight threads: the refactor converts "warn-but-accept" into "hard-reject via ValidationError." By design — previous loose behavior is what allowed the stringified-JSON crash #153 worked around in the frontend.
- `manager.py` helpers still produce dicts with stripped fields. Zero callers, so nothing breaks; flagged for follow-up cleanup PR.

## Test plan

- [x] 27 tool-level tests pass.
- [x] 13 middleware tests pass (ported + extended from #156, +2 for #160).
- [x] Full unit suite: 2784/2784 pass.
- [x] Ruff: clean.
- [x] Pydantic JSON schema verified — LLM sees `{content, activeForm, status}` only.
- [x] Backward-compat verified — extra fields silently ignored.
- [x] Middleware payload normalization verified — uppercase status → lowercase at frontend; legacy fields stripped.
- [x] Failed-event branch verified — None/int payloads no longer crash in secondary TypeError.
- [x] Failed-event payload verified normalized — same whitelist + lowercase as success path (#160).

### DIFF
--- a/src/ptc_agent/agent/middleware/todo_operations/sse_middleware.py
+++ b/src/ptc_agent/agent/middleware/todo_operations/sse_middleware.py
@@ -86,10 +86,12 @@ class TodoWriteMiddleware(AgentMiddleware):
         try:
             result = await handler(request)
 
-            # Calculate status counts
+            # Calculate status counts. Lowercase to mirror the TodoItem
+            # validator's normalize_status — an LLM-sent "PENDING" would
+            # otherwise fall through as uncounted.
             status_counts = {"pending": 0, "in_progress": 0, "completed": 0}
             for todo in todos:
-                status = todo.get("status", "unknown")
+                status = str(todo.get("status", "")).lower()
                 if status in status_counts:
                     status_counts[status] += 1
 
--- a/src/ptc_agent/agent/middleware/todo_operations/sse_middleware.py
+++ b/src/ptc_agent/agent/middleware/todo_operations/sse_middleware.py
@@ -96,27 +96,29 @@ class TodoWriteMiddleware(AgentMiddleware):
             # Continue with tool execution even if streaming fails
             return await handler(request)
 
+        # Restrict payload to the TodoItem schema: strip any legacy
+        # or unknown fields the LLM may still send, and lowercase status
+        # so the frontend's strict equality (status === 'pending') holds
+        # regardless of case variance. Anything non-dict is dropped.
+        # Computed before the try block so the failed-event path below
+        # ships the same normalized shape instead of raw LLM input.
+        status_counts = {"pending": 0, "in_progress": 0, "completed": 0}
+        normalized_todos = []
+        for todo in todos:
+            if not isinstance(todo, dict):
+                continue
+            status = str(todo.get("status", "")).lower()
+            if status in status_counts:
+                status_counts[status] += 1
+            normalized_todos.append({
+                "content": todo.get("content", ""),
+                "activeForm": todo.get("activeForm", ""),
+                "status": status,
+            })
+
         # Execute the actual tool
         try:
             result = await handler(request)
-
-            # Restrict payload to the TodoItem schema: strip any legacy
-            # or unknown fields the LLM may still send, and lowercase status
-            # so the frontend's strict equality (status === 'pending') holds
-            # regardless of case variance. Anything non-dict is dropped.
-            status_counts = {"pending": 0, "in_progress": 0, "completed": 0}
-            normalized_todos = []
-            for todo in todos:
-                if not isinstance(todo, dict):
-                    continue
-                status = str(todo.get("status", "")).lower()
-                if status in status_counts:
-                    status_counts[status] += 1
-                normalized_todos.append({
-                    "content": todo.get("content", ""),
-                    "activeForm": todo.get("activeForm", ""),
-                    "status": status,
-                })
 
             # Build completed event with structure expected by streaming_handler
             # Must include artifact_type for handler to recognize and emit as SSE artifact event
@@ -166,8 +168,11 @@ class TodoWriteMiddleware(AgentMiddleware):
                 "timestamp": failed_timestamp,
                 "status": "failed",
                 "payload": {
-                    "todos": todos,
-                    "total": len(todos),
+                    "todos": normalized_todos,
+                    "total": len(normalized_todos),
+                    "completed": status_counts["completed"],
+                    "in_progress": status_counts["in_progress"],
+                    "pending": status_counts["pending"],
                     "error": str(e),
                 },
             }
--- a/src/ptc_agent/agent/middleware/todo_operations/sse_middleware.py
+++ b/src/ptc_agent/agent/middleware/todo_operations/sse_middleware.py
@@ -67,7 +67,21 @@ class TodoWriteMiddleware(AgentMiddleware):
 
         tool_call_id = tool_call.get("id", "unknown")
         tool_args = tool_call.get("args", {})
-        todos = tool_args.get("todos", [])
+        raw_todos = tool_args.get("todos", [])
+
+        # Defensive: the tool signature now enforces List[TodoItem] via pydantic,
+        # so a valid handler return implies a proper list. But the failed-event
+        # branch below calls len(todos) on the raw LLM input — a None or int
+        # payload would crash the error path and swallow the original
+        # ValidationError. Coerce to [] so both branches are safe.
+        if isinstance(raw_todos, list):
+            todos = raw_todos
+        else:
+            logger.warning(
+                f"[TODO_MIDDLEWARE] Non-list todos payload "
+                f"(type={type(raw_todos).__name__}); coercing to []"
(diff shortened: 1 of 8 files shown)


Chen-zexi:
False positive — the concern misses the existing `Array.isArray` coercion at the top of each handler.

Both handlers normalize `payload.todos` to an array **before** the assignment lines I touched:

- [`streamEventHandlers.ts:545`](https://github.com/ginlix-ai/LangAlpha/blob/9c41d3ec311945f2c4bbbef26a2e4fcccd075437/web/src/pages/ChatAgent/hooks/utils/streamEventHandlers.ts#L545) — `const todos = Array.isArray(payload.todos) ? payload.todos : [];`
- [`historyEventHandlers.ts:702`](https://github.com/ginlix-ai/LangAlpha/blob/9c41d3ec311945f2c4bbbef26a2e4fcccd075437/web/src/pages/ChatAgent/hooks/utils/historyEventHandlers.ts#L702) — same pattern.
- Triple-safety at the leaf in [`TodoDrawer.tsx:61`](https://github.com/ginlix-ai/LangAlpha/blob/9c41d3ec311945f2c4bbbef26a2e4fcccd075437/web/src/pages/ChatAgent/components/TodoDrawer.tsx#L61).

So even if a pre-migration stored event has `todos: null` in `conversation_responses.sse_events`, `todos` is already `[]` by the time it reaches the spots I edited. The `|| []` fallback was tautological in **both** stream and history paths, not just the stream path.

The split reasoning (safe in stream via middleware normalization, unsafe in history via stored events) only holds if you ignore the upstream `Array.isArray` guard, which protects both equally. The `??` vs `||` nuance is also moot at this specific assignment site because the value is guaranteed-array before it gets here.

Chen-zexi:
Both observations valid, no code changes needed.

**#160 closure** — the PR body already carries `Closes #160` (added with the `bbce68c9` fix), so merge will auto-close it.

**`isinstance(todo, dict)` is load-bearing** — agreed, and this is a correction to the earlier bot review on this PR which called the check "unreachable on the success path." After `bbce68c9` lifted the normalization above `await handler(request)`, the loop runs on raw LLM dicts pre-validation. Non-dict items are fully reachable: LLM sends mixed list → isinstance filter drops garbage → pydantic rejects the whole call → failed-event branch ships the already-normalized `normalized_todos`. The in-code comment at `sse_middleware.py:99-104` ("Anything non-dict is dropped" + "Computed before the try block so the failed-event path below ships the same normalized shape") already documents this correctly.

PR #176, opened 2026-04-25
feat(memo): user-managed document memo store

Chen-zexi:
## Summary

Ships **Step 2** of the long-term memory plan: an **investment-memo store** that closes the loop **research task → agent output → iteration → finalized memo → reusable knowledge base**. Memos are user-managed deliverables (theses, analyses, recommendations, source PDFs / SEC filings / earnings decks) that follow the user across every workspace and thread.

Built on the same primitive as long-term memory: `CompositeFilesystemBackend` routes `.agents/user/memo/**` to a read-only `StoreBackend` over the LangGraph `BaseStore` (`AsyncPostgresStore` in production), while everything else stays on the ephemeral Daytona sandbox. That's the **cross-sandbox solution** — sandbox filesystems die with each workspace, but the store outlives them, so a memo finalized in one workspace is visible from every other workspace's agent on the next `read_file` / `glob`.

Two ingest paths: upload from disk (markdown / PDF / text / CSV / JSON), **or** promote a sandbox-produced file directly via "Add to memo" — the sandbox path records `source_workspace_id` + `source_path` so re-promoting the same file dedups in place instead of fragmenting the thesis. PDFs are text-extracted server-side; original bytes land in S3-compatible object storage (R2 / COS / MinIO / S3) with base64 fallback. Description + summary are generated asynchronously through `LLMService` (BYOK / OAuth-correct). The agent gets read-only access at `.agents/user/memo/` and is told the memo count via a tiny `<memo-index count="N"/>` block injected after the prompt-cache breakpoint — discoverable without leaking titles into every prompt.

The 12 commits are organized as 3 prep refactors → 3 service primitives → 4 memo feature layers → 1 docs commit → 1 perf optimization. Each commit is bisectable on its own.

### Backend prep

- **`refactor(backends): rename memory.py → langgraph_store.py, add read-only flag`** — generalizes the store-backed filesystem from memory-only to any prefix route. `ReadOnlyStoreError` enables the read-only memo tier.
- **`perf(store): add per-request store cache`** — K model calls in a turn previously triggered K `memory.md` aget()s. The cache memoizes one read per `(namespace, key)`; backend invalidates on writes.
- **`refactor(server): extract _store_helpers`** — hoists `asearch`/`aget`/`validate_key`/`paginate` helpers out of the memory router so the new memo router can share them.
- **`fix(api_call): flatten content_blocks before manual JSON parse`** — when `make_api_call` falls back to scanning `.content` for JSON, models that return `[reasoning_block, text_block]` lists bypassed the regex. Flatten via `format_llm_content` first.

### Service primitives

- **`feat(llm): add generic one-shot LLMService wrapper`** — wraps `resolve_llm_config` + `make_api_call` so every server-side utility LLM call (memo metadata, future hint / title / follow-up generators) shares one BYOK / OAuth-aware entry point. Logs at INFO when falling back to the platform key. Optional structured response schema.
- **`feat(storage): add get_bytes + cache the boto3 client`** — both providers now expose `get_bytes` for reading objects (previously write-only). Lazy-cached boto3 client reuses TLS sessions.

### Memo feature

- **`feat(memo): user-managed document store backend + HTTP API`** — `.agents/user/memo/`, slug + collision-suffix keys, pdfplumber → pypdf text extraction, async LLM metadata, deterministic memo.md regeneration, R2/S3/COS binary storage with base64 fallback. 7-endpoint router (upload / list / read / write / delete / download / regenerate). Includes review / adversarial-pass hardening: workspace ownership check on sandbox sources, source_path length cap, chunked upload reader, MAX_LIST_LIMIT cap on dedup walk, write/regenerate handlers wrap RMW in `lock_for_namespace`, `_merge_metadata` always re-fetches the row before merging, RFC 6266-compliant Content-Disposition for CJK / emoji filenames, MIME fallback to extension when uploads report `application/octet-stream` (Safari + drag-drop).
- **`feat(memo): agent integration — middleware, prompt, tool guards`** — `MemoAwarenessMiddleware` injects `<memo-index count=N/>` after the prompt cache breakpoint; static `memo.md.j2` describes the read-only contract once at the cacheable prefix. `bash` and `execute_code` refuse memo paths; `file_ops Read` bypasses the multimodal branch for memo PDFs (text lives in the store). Memo-path detection accepts `./.agents/`, `/.agents/`, and bare `.agents/`.
- **`feat(agent,server): wire memo route + LLMService into runtime`** — agent factory adds the third store-backed route and shares one `RequestScopedStoreCache`. Server lifespan constructs `LLMService` once and registers the memo router.
- **`feat(web): memo panel UI + tests`** — full CRUD: drag-drop upload, list with pending / ready / failed status, markdown / text / PDF viewer, text editor, regenerate + delete + bulk delete. FilePanel surfaces a "memoed" indicator on workspace files. 40 vitest cases.

### Docs

- **`docs: document memo store + agent middleware updates`** — `README.md`, `CLAUDE.md`, `web/CLAUDE.md`, `web/README.md`.

## Test Coverage

```
Backend: 3068 tests passing (memo backend ~82% line coverage)
Frontend: 740 tests passing (40 new for memo)
Combined coverage on the 1820 LOC of new frontend memo code: ~75% of regression-critical UI flows
```

Coverage map:

```
Upload → text + PDF (page-cap, scanned-PDF reject, slug collision, sandbox dedup, ownership) ★★★
Write  → 404, 400-binary, 413-too-large, sha256 short-circuit, pending kickoff             ★★★
List/Read/Download → R2 + base64 + ASCII-fallback header + RFC 6266 + injection defense     ★★★
Delete → cascades memo.md rebuild, cancels pending metadata task                            ★★★
Background metadata → ready / failed / mid-flight modify / mid-flight delete                 ★★★
Agent path → read_file, write_file rejected, glob/grep merge, bash/exec rejection           ★★★
Middleware → count injection, 500+ display cap, timeout/error skip, cache dedup              ★★★
Frontend  → list/empty/viewer/edit-save/delete/regenerate/upload + PDF blob                 ★★★
```

## Pre-Landing Review

11 issues (3 critical, 8 informational). Resolved:
- Workspace ownership check, source_path cap, chunked upload, metadata length caps, `_find_by_source` walk cap, bulk-delete N+1, LLMService platform-fallback log, `memory_enabled` / `memo_enabled` threading.

## Adversarial Review

Claude + Codex passes. Codex P1 (metadata CAS resurrects deleted/edited rows) + Codex P2s (memo path leading slash, MIME fallback) all fixed. Claude's lock-on-write/regenerate fixed. Defer: boto3 timeouts, second-resolution timestamp CAS, `rebuild_memo_index` lock — see Follow-ups.

## Documentation

`docs:` commit lands updates to:
- **`README.md`** — middleware count 24 → 25, persistent-workspaces section mentions memo.
- **`CLAUDE.md`** — new Long-Term Memory & Memo Store subsection (user/workspace tiers, S3-compatible binary storage, request-scoped cache); Server Utility LLM Calls subsection on `LLMService.complete`.
- **`web/CLAUDE.md`** — File uploads (memo) paragraph in Data Fetching.
- **`web/README.md`** — Memo Library bullet.

## Follow-ups (not in scope)

- Lock-split: drop the namespace lock during R2 PUT, re-acquire for the binary_ref aput. Today the lock is held during R2 upload, so a slow PDF upload serializes every other memo op for that user (~3-5s). Mitigated by the dedup walk cap; deferred because it needs careful CAS design.
- boto3 `connect_timeout` / `read_timeout` on `asyncio.to_thread` calls.
- Microsecond-resolution `modified_at` for tighter CAS in burst regen.
- `_METADATA_TASKS` registry → Redis or store-backed for multi-worker deployments.
- `rebuild_memo_index` lock or per-user catalog-row CAS to prevent the (rare) race where a metadata-finish merge resurrects a deleted memo's catalog entry.

## Test plan

- [x] All backend pytest unit tests pass (3068, 0 failures)
- [x] All frontend vitest tests pass (740, 0 failures)
- [x] Lint clean on the diff scope (74 ruff errors elsewhere are pre-existing)
- [x] Manual smoke: drag a markdown file into MemoPanel → "Summary generating…" → real description within ~5s
- [x] Manual smoke: upload a PDF → text extracted, agent can `read_file` it, original bytes downloadable
- [x] Manual smoke: try to `write_file` to a memo path from chat → agent gets read-only error
- [x] Manual smoke: upload a CJK-named file (`备忘录.pdf`) → download header survives Starlette latin-1
- [x] Manual smoke: re-upload identical content → no second LLM call (sha256 short-circuit)
- [x] Manual smoke: switch to a new workspace → same memos still present (user-scoped persistence)


Chen-zexi:
@claude — thanks for the review. Folded all three fixes into commit `2fb478af` (the original `feat(memo): user-managed document store backend + HTTP API`) via `--fixup` + `--autosquash` so the merge PR keeps one self-contained memo commit instead of growing a fix-on-fix tail. Intent: bisectability + clean history at merge time.

### Blocking — `delete_user_memo` namespace lock

Wrapped aget + `_cancel_pending_metadata` + adelete in `async with lock_for_namespace(namespace):`. S3 `delete_binary` and `rebuild_memo_index` stay outside the lock (best-effort, eventual). Added `test_delete_acquires_namespace_lock` (`tests/unit/server/app/test_memo_api.py`) — holds the registry-shared lock externally, fires delete in a background task, asserts the task does not complete until the lock is released. Fails on `main` without the fix; passes after.

### Non-blocking — stale `_merge_metadata` comment

Removed the dead `base_value` parameter from `_merge_metadata` entirely (only call site already in this PR; no external callers) plus the stale "skip the second aget" comment at the call site. Updated the docstring so the CAS-via-fresh-aget contract is the only described behavior.

### Non-blocking — unguarded `base64.b64decode`

Wrapped in `try/except (binascii.Error, ValueError)` → `HTTPException(500)` with `logger.exception`, mirroring the `fetch_binary` failure path immediately above it. Added `import binascii`.

### Deferred (still in Follow-ups, unchanged)

- Lock-split during R2 PUT.
- `_METADATA_TASKS` Redis-backed for multi-worker.

Verification: 3069 unit tests pass, including the new regression test. Lint clean on diff. Could you confirm the amended commit closes the blocker as intended?

PR #177, opened 2026-04-26
fix(model-picker): show full accessible list + hide shadowed built-ins

Chen-zexi:
## Summary

Two related repairs to the model picker:

**1. Unscope Settings + Setup-wizard dropdowns from the starred filter.** The Primary/Flash dropdowns in Settings → Model and the Setup wizard's DefaultsStep now show every model the user has access to (post tier / BYOK / OAuth gating), not just the starred subset. The starred-models filter stays on the chat-input picker only, where vertical space is tight. PR #164 (\`77bc9fde\`) had funneled both surfaces through \`useStarredPickerModels\` and conflated two concerns under one preference key — chat-input quick-access vs. full picker scope. The Settings chip-strip below the dropdown is the management UI for the chat-input list; it shouldn't double as a gate on the dropdown above it.

**2. Hide built-ins shadowed by a custom_models entry.** When a user adds a custom_models entry that shares a name with a built-in, the resolver's \`classify_model\` returns CUSTOM first (\`src/server/handlers/chat/llm_config.py:455-483\`) — the platform path becomes unreachable. The picker was still showing the built-in row under its parent group with a misleading \`(platform)\` badge, producing duplicate rows that routed identically. \`buildVisibleModels\` now drops shadowed built-ins from non-owner groups (preserved in \`rawModels\` so ModelPickStep still shows them under the parent catalog when the user is configuring it directly). \`useModelAccessMap\` resolves custom-pair rows against the owning group key, so they badge as \`byok\` instead of falling back to the built-in's \`meta.provider\`.

## Files

- \`web/src/pages/Settings/Settings.tsx\`, \`web/src/pages/Setup/steps/DefaultsStep.tsx\` — drop \`useStarredPickerModels\` import + derivation, swap prop to \`visibleModels\`/\`models\`
- \`web/src/hooks/useFilteredModels.ts\` — shadow-hide step in \`buildVisibleModels\`, exposes \`customPairs\` on \`BuildVisibleModelsResult\`. Uses \`Object.hasOwn(shadowOwner, m)\` for the lookup (defensive against \`Object.prototype\` name collisions)
- \`web/src/hooks/usePlatformModels.ts\` — \`useModelAccessMap\` accepts \`customPairs\`, resolves provider via \`groupKey\` for custom-pair rows
- \`web/src/hooks/useAllModels.ts\` — wires \`customPairs\` through, adds it to the loading-state default
- \`web/src/hooks/useStarredPickerModels.ts\` + test — deleted (zero remaining callers)
- \`web/src/hooks/__tests__/useFilteredModels.test.ts\`, \`usePlatformModels.test.ts\` — added shadow-hide and badge-map tests, neutralized to generic placeholder names

## Test plan

- [x] \`pnpm vitest run src/hooks/__tests__/useFilteredModels.test.ts src/hooks/__tests__/usePlatformModels.test.ts\` — 41/41 pass
- [x] \`pnpm lint\` — 0 errors (94 pre-existing warnings in untouched files)
- [ ] Manual: Settings → Model tab dropdown shows full tier-accessible list
- [ ] Manual: Setup wizard Primary/Flash dropdowns show full list
- [ ] Manual: Chat-input model dropdown still scoped to starred set (regression watch)
- [ ] Manual: Add a custom_models entry shadowing a built-in name, verify duplicate row is gone and badge is \`byok\`


Chen-zexi:
Good catch — applied in 1dbde2d1.

```diff
- if (!shadowOwner[cm.name]) shadowOwner[cm.name] = cm.provider;
+ if (!Object.hasOwn(shadowOwner, cm.name)) shadowOwner[cm.name] = cm.provider;
```

Closes both the symmetry mismatch and the empty-string-provider edge case. 41/41 tests still pass.

PR #178, opened 2026-04-26
fix(memo): hardening + cross-worker metadata cancel

Chen-zexi:
## Summary

Six follow-ups on top of the memo feature shipped in #176. Five from the original hardening plan, plus one bug a smoke-test caught.

**Performance**
- **Lock-split upload.** Extract + R2 PUT run outside the per-namespace lock; only the catalog write stays inside. Uploads no longer block other memo ops on the same user for the 3-5 s the S3 round trip takes. Storage keys are UUIDs, so slug collisions can't overwrite each other's bytes.

**Robustness**
- **Microsecond `modified_at`.** Burst regenerate (5 clicks in one wall-clock second) used to produce identical second-precision timestamps, defeating the CAS check.
- **Lock around `rebuild_memo_index`.** All 7 call sites now serialize. Closes a resurrection race where concurrent rebuilds could interleave `asearch` + `aput` and leave `memo.md` claiming a deleted memo still exists.
- **Cross-worker metadata cancel.** Redis carries a cooperative cancel signal across uvicorn workers. A delete on worker A can now stop a pending LLM call on worker B. Single-process dev unaffected. Fail-open on Redis outage.
- **boto3 timeouts.** 5 s connect / 30 s read instead of botocore's 60/60 default. Configurable via env vars.
- **NUL byte sanitize.** Strip `\x00` at the three memo content ingestion points. `pdfminer` can emit NUL from malformed font glyphs and Postgres JSONB rejects it; smoke-caught on a real Android lecture PDF that had been failing 503.

**Cleaner error codes**
- Storage upload failures: 502 (was 500).
- Store-write failures during upload: 503 with retry hint (was 500).

## Test Coverage

117 unit tests pass across `tests/unit/server/app/test_memo_api.py`, `tests/unit/ptc_agent/agent/memo/`, `tests/unit/utils/storage/`, `tests/unit/server/services/`. New tests cover:

- boto3 timeout config + e2e upload→502 mapping for ConnectTimeoutError / ReadTimeoutError
- Microsecond `now_iso()` uniqueness + legacy second-precision CAS coexistence
- Concurrent delete during merge does not resurrect row
- Phase A → Phase B race + orphan blob cleanup
- Cross-worker handover SET → DELETE → SET ordering
- Cross-worker cancel before + after LLM call
- NUL byte stripping on PDF extract, UTF-8 decode, and `write_memo` paths
- Backwards compat: pre-PR slug-derived `binary_ref.key` still downloads cleanly

## Pre-Landing Review

Already CLEAR via prior `/review` run on this work. Specialists dispatched: testing, maintainability, security, performance, api-contract. PR Quality Score: 7.5. All critical findings either fixed or explicitly skipped (1 deferred — inflight-key endpoint, documented as follow-up). Adversarial review (Claude + Codex) passed.

## Plan Completion

All 5 plan items DONE plus 1 smoke-caught fix. The deferred item is the inflight-key GET endpoint for cross-worker UI visibility, documented inline as a known limitation.

## Smoke Test

Manually verified:
- Upload of a real PDF that previously crashed with 503 (pdfminer NUL bytes) now succeeds.
- Existing pre-PR memos still download correctly (slug-derived `binary_ref.key` backwards compat).

Recommend two minutes of clicking before merge:
- Open one pre-PR memo's preview to confirm legacy keys load.
- Upload a fresh PDF and watch the description/summary land.

## Test plan

- [x] All memo-related unit tests pass (117 / 117)
- [x] Ruff clean on changed files
- [x] No SQL injection / unsafe deserialization
- [ ] Smoke: open pre-PR memo preview
- [ ] Smoke: upload fresh PDF, confirm metadata renders

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -109,6 +109,8 @@ redis:
     workflow_status: 3600   # Completed/cancelled workflow status key TTL (1 hour)
     cancel_flag: 300        # Workflow cancel flag TTL (5 minutes)
     steering: 3600          # Steering message key TTL (1 hour)
+    memo_metadata_inflight: 300  # Cross-worker visibility of in-flight memo metadata tasks (5 minutes)
+    memo_metadata_cancel: 60     # Cooperative cross-worker cancel flag for memo metadata (1 minute)
 
     # Per-interval OHLCV cache TTLs (≈ interval_seconds × 1.5)
     # SWR soft threshold stays at 50% of these values
--- a/src/config/models.py
+++ b/src/config/models.py
@@ -106,6 +106,19 @@ class RedisTTLConfig(BaseModel):
     steering: int = Field(
         default=3600, description="TTL for steering message Redis keys (1 hour)"
     )
+    memo_metadata_inflight: int = Field(
+        default=300,
+        description=(
+            "TTL for the cross-worker visibility key marking a memo metadata "
+            "task as in flight (5 minutes)"
+        ),
+    )
+    memo_metadata_cancel: int = Field(
+        default=60,
+        description=(
+            "TTL for the cooperative cross-worker memo metadata cancel flag (1 minute)"
+        ),
+    )
 
 
 class RedisSWRConfig(BaseModel):
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -290,6 +290,16 @@ def get_redis_ttl_steering() -> int:
     return get_infrastructure_config().redis.ttl.steering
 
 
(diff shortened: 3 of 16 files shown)


Chen-zexi:
We'll do a thorough import-consistency audit across the repo in a separate PR

All other findings from the three review rounds are addressed in `c45e3227` (await handover) and `70f5922c` (fail-closed handover + `_rebuild_index_under_lock` consistency). Treating as ready to merge.

PR #184, opened 2026-04-28
fix(sandbox): consolidate Playwright Chromium install for Scrapling

### LINKED ISSUE #149: bug: Playwright Chromium binary missing for Scrapling sessions
## What I was doing
Running a Python smoke test for MCP tools that exercise the Scrapling browser session APIs.

## What went wrong
`scrapling.open_session` failed with a missing executable error:

`BrowserType.launch_persistent_context: Executable doesn't exist at /usr/local/ms-playwright/chromium-1208/chrome-linux64/chrome`

## Where the issue likely is
Sandbox/browser provisioning for Playwright-based tools, especially Scrapling session startup.
The expected Chromium path does not exist in the environment.

## Suggested fix
Ensure Playwright browser binaries are installed and available in the runtime image, or add a startup/bootstrap step that runs the browser install for Chromium before Scrapling session use.

## Environment
- Tool/Skill: MCP smoke test via Python
- Component: Scrapling `open_session`
- Error type: sandbox

@claude Please triage this issue — verify the root cause, assess severity, and suggest a fix if straightforward.

mvanhorn:
## Summary

The sandbox image installs Chromium twice via two different package managers, which can pin to different Chromium versions. Drop the npm-side install so the Python `playwright` package (used by Scrapling at runtime) is the single source of truth.

## Why this matters

Issue [#149](https://github.com/ginlix-ai/langalpha/issues/149) was filed by @langalpha-bot:

> `scrapling.open_session` failed with a missing executable error:
> `BrowserType.launch_persistent_context: Executable doesn't exist at /usr/local/ms-playwright/chromium-1208/chrome-linux64/chrome`

Walking the Dockerfile.sandbox layout:
- Lines 41-43: `npm install -g playwright` then `npx playwright install --with-deps chromium`
- Line 75: `ENV PLAYWRIGHT_BROWSERS_PATH=/usr/local/ms-playwright`
- Lines 93-95: `uv pip install ... "scrapling[all]" playwright`
- Line 101: `RUN scrapling install` (which runs `python -m playwright install --with-deps chromium`)

The npm `playwright` package and the Python `playwright` package can pin to different Chromium versions. At runtime, scrapling launches sessions via Python `playwright`, which expects the version it pinned to. If only the npm-pinned version landed in `/usr/local/ms-playwright/chromium-NNNN/`, scrapling raises the missing-executable error.

The runtime sandbox does not invoke `npx`, only Python `playwright`. The npm install path was redundant.

## Approach

Drop the npm-side Playwright. Keep `docx` and `pptxgenjs` as npm globals because they are used as npm CLIs by the sandbox. Make `scrapling install` the single Chromium source. Refresh the stale comment that said browsers downloaded to `/root/.cache/` (they go to `/usr/local/ms-playwright/` per the ENV var).

## Changes

`Dockerfile.sandbox`:
- Line 41: removed `playwright` from the npm install list, kept `docx` and `pptxgenjs`.
- Lines 42-43: removed `npx playwright install --with-deps chromium` (now handled by `scrapling install`).
- Line 101: `RUN PLAYWRIGHT_BROWSERS_PATH=/usr/local/ms-playwright scrapling install` (belt-and-suspenders, the ENV var is also set above).
- Comment refresh on the cache-copy block to reflect the actual layout.

Net diff: +11/-9.

## Testing

This fix changes the sandbox runtime image. There is no Dockerfile-build CI gate in the repo (CI is Python-side). I did not run `docker build` locally as part of this PR. The Python lint (`uv run ruff check src/`) is unaffected.

Recommended local validation before merge:

```
docker build -f Dockerfile.sandbox -t langalpha-sandbox-test .
docker run --rm langalpha-sandbox-test python3 -c "
from scrapling.fetchers import open_session
import asyncio
async def main():
    async with await open_session(headless=True) as session:
        page = session.page
        await page.goto('https://example.com')
        print(await page.title())
asyncio.run(main())
"
```

If the smoke test passes (Chromium binary exists at the path scrapling expects), the fix is good.

Closes #149.

### DIFF
--- a/Dockerfile.sandbox
+++ b/Dockerfile.sandbox
@@ -38,9 +38,7 @@ RUN apt-get update \
     && tar -xJf /tmp/node.tar.xz -C /usr/local --strip-components=1 \
     && rm /tmp/node.tar.xz \
     # -- NPM global packages --
-    && npm install -g playwright docx pptxgenjs \
-    && PLAYWRIGHT_BROWSERS_PATH=/usr/local/ms-playwright \
-       npx playwright install --with-deps chromium \
+    && npm install -g docx pptxgenjs \
     # -- GitHub CLI (multi-arch: amd64 / arm64) --
     && GH_ARCH=$(dpkg --print-architecture) \
     && curl -fsSL "https://github.com/cli/cli/releases/download/v2.87.3/gh_2.87.3_linux_${GH_ARCH}.tar.gz" \
@@ -69,7 +67,7 @@ RUN apt-get update \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-# Playwright browsers were installed to /usr/local/ms-playwright (as root).
+# Playwright browsers are installed to /usr/local/ms-playwright (as root).
 # Set the env var so the Python playwright package finds them at runtime
 # regardless of which user runs the sandbox.
 ENV PLAYWRIGHT_BROWSERS_PATH=/usr/local/ms-playwright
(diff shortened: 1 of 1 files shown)


Chen-zexi:
Thanks @mvanhorn — diagnosis is correct: npm and Python `playwright` pin different Chromium versions, scrapling drives the Python one at runtime, so it looked for `chromium-1208` where only the npm version landed.

Two reasons we went a different way:

1. **Scope.** #149 came from the production runtime, which builds its sandbox from the Daytona snapshot in `src/ptc_agent/core/sandbox/providers/daytona.py` — a separate image definition with the same split. This PR only touches `Dockerfile.sandbox`, so the environment that actually failed would stay broken.
2. **The npm side isn't redundant.** The agent can drive node/npm `playwright` from its own code to test frontends it builds (e.g. a dashboard it just generated). Our source never calls `npx playwright`, but the sandbox runs agent-authored code, so dropping it removes a real capability.

#250 fixes the split without losing that: the npm and Python installs share one `PLAYWRIGHT_BROWSERS_PATH`, applied in both definitions so they can't drift again. Closing as superseded by #250 — appreciate the PR.

PR #185, opened 2026-04-28
feat(subagent): bound per-task heap, durable Redis-spilled SSE history

Chen-zexi:
## Summary

Reduces resident memory for langalpha workers handling parallel subagent
fan-out and fixes three correctness bugs uncovered along the way. The
in-memory state held by completed background tasks could grow without
bound, the LangSmith tracer plumbing was racing under concurrent fan-out,
and a stale UI gate left finished subagents showing "Running" until all
siblings closed. Five logical commits, each independently bisectable.

## What was leaking

- **`BackgroundTask` heap.** Completed entries retained the compiled
  LangGraph, every middleware closure, the completion callback, asyncio
  task handles, and metadata closures (`handler`, `sandbox`,
  `token_callback`) for the full 24h `result_ttl`. Stair-stepped resident
  memory under sustained fan-out; never reclaimed.
- **`captured_events`.** Unbounded `list[dict]` per subagent — one Python
  dict per emitted event, retained for the lifetime of the thread. A
  single fan-out could pin 100k+ event objects; concurrent fan-outs
  dominated worker heap.
- **Shared `LangChainTracer`.** Concurrent subagents inherited
  `parent_config["callbacks"]` containing one shared tracer instance. Its
  mutable `run_map`/`order_map` raced under fan-out, leaking tracker
  state into the parent middleware lifetime and producing collapsed run
  trees in LangSmith.
- **Custom LangSmith retry wrapper.** Retried on every error class;
  amplified a single failed upload into many redundant attempts, holding
  payload state until exhausted.

### `feat(background-task)` — heap bounds + collector helpers
- `TaskInfo.result_buffer` is a `deque(maxlen=1000)` instead of an
  unbounded list.
- `_release_terminal_refs` runs after persistence on every terminal mark
  (`_mark_completed`/`_failed`/`_cancelled`/`_soft_interrupted`). Drops
  the heavy refs while preserving small scalars (`task.result`,
  `task.completed`, `task.task_id`, `task.display_id`) so cross-turn
  `TaskOutput(task_id="…")` keeps working.
- `_await_drain_and_cleanup_tasks` drops `asyncio_task`/`handler_task`
  and clears `new_event_signal` once the SSE drain finishes.
- New `iter_subagent_events_full` collector joins a Redis-spilled prefix
  with the in-memory tail at a stable high-water mark. Emits a structured
  `subagent_history_truncated` warning when a gap can't be recovered, so
  silent truncation in `conversation_responses.sse_events` is observable
  instead of shipping incomplete history.

### `feat(subagent)` — Redis-spilled `captured_events` + bounded SSE reconnect
- `captured_events` (unbounded list) → `captured_events_tail` (deque
  capped at the documented reconnect-buffer size).
- Producer-side Redis spool keyed by
  `subagent:events:{thread_id}:{task_id}` via `pipelined_event_buffer`.
- 500 ms `asyncio.wait_for` cap on every spill. Timeout or pipeline error
  flips a sticky `redis_write_failed` flag so subsequent appends
  short-circuit — Redis latency cannot pace agent execution on the hot
  path.
- `captured_events_warn_threshold` emits a structured high-water log.
- `stream_subagent_task_events` drains the tail by seq cursor and falls
  back to `iter_subagent_events_full` for any gap below `tail_front_seq`,
  including reconnect (which reads the durable Redis spool from seq 1).
- `middleware._reset_task_for_resume` clears both Redis spool keys
  (events list + meta hash) before resetting task fields so a re-run
  doesn't interleave with prior-run records.

### `feat(subagent)` — per-tracer isolation + drop parent callbacks
- Strip `parent_config["callbacks"]` before invoking the subagent; pass a
  fresh filtered LangSmith tracer per call.
- Per-call token tracker reattached via
  `current_background_token_tracker` so accounting rolls up to the
  parent.
- Rename resume helpers (`_get_background_task_context`,
  `has_bg_task_context`, `bg_checkpoint_ns`, `invoke_state`) so call
  sites read as "invoke at a known checkpoint_ns" rather than "resume a
  paused run."

### `refactor(langsmith)` — drop custom trace filter
- Delete `src/config/langsmith.py` (custom `strip_base64_from_trace`
  anonymizer + `get_filtered_langsmith_tracer` factory) and its tests.
- `_common.py` no longer constructs a filtered tracer for chat
  callbacks.
- The SDK's built-in retry policy (3 attempts on connection errors only)
  is the right level for transient upload failures; the per-tracer
  isolation in the prior commit removes the shared-state fan-out that
  originally motivated the wrapper.
- Net: 163 lines of custom anonymizer + 215 lines of test deleted.

### `fix(chat-agent)` — flip subagent chat-card status on each per-task SSE close
- `markAllSubagentTasksCompleted` was gated on
  `subagentStreamsRef.current.size === 0`, so chat-card status only
  flipped when the LAST per-task stream closed. Mid-flight, a finished
  subagent showed "Running" in the chat view while the left-panel
  timeline already read "Completed."
- Move the call out of the size-zero branch. The function is idempotent
  and skips active-stream tasks, so calling on every per-task close
  flips only the just-finished one.

## Outcome

- Resident memory stays flat across parallel subagent fan-out cycles
  instead of growing monotonically; OOM under sustained load is
  resolved.
- SSE reconnect within the workflow's lifetime returns the full event
  history when Redis is healthy, and surfaces a structured warning when
  it can't.
- Concurrent subagents have isolated tracer state — properly nested run
  trees in LangSmith, no cross-subagent leak into parent middleware.
- Chat-card and timeline panels agree on subagent status at every
  transition.

## Test plan
- [x] `uv run pytest tests/unit/middleware/ tests/unit/server/services/ tests/unit/server/handlers/` — 180 pass
- [x] `cd web && pnpm vitest run src/pages/ChatAgent/hooks` — 69 pass
- [x] Coverage exercises: terminal cleanup on all four mark paths;
      Redis spill happy-path / timeout / circuit-breaker; reconnect cold
      / mid-stream / malformed record; per-subagent tracer isolation
      regression; resume cleanup of Redis spool keys; truncation
      warning.
- [x] Pre-landing review: 0 critical, 5 informational (3 auto-fixed and
      squashed into the right commits, 2 deferred).


Chen-zexi:
Re #2 (`_release_terminal_refs` clearing `result_buffer`) — this one's architecturally inert in practice.

The frontend reconnect protocol probes `/status` *before* deciding which endpoint to use:

- `useChatMessages.ts:2088` (retry loop), `:2147` (report-back), `:2190` (history load) all call `getWorkflowStatus(threadId)` first.
- `/messages/stream` (in-memory + Redis) is only called when `status.can_reconnect === true`.
- `workflow_handler.py:219` defines `can_reconnect = status in [ACTIVE, DISCONNECTED]` — `COMPLETED`/`FAILED`/`CANCELLED` all return `false`.
- After `can_reconnect: false` the frontend falls through to `replayThreadHistory` → `/messages/replay`, which reads from Postgres `conversation_responses.sse_events` (the durable replay source, 24h TTL).

So the only path where `result_buffer` is consulted is mid-flight reconnect (status = ACTIVE/DISCONNECTED), and on that path `_release_terminal_refs` hasn't fired yet. Once the workflow hits a terminal state, `/status` flips `can_reconnect` to `false`, the frontend routes to Postgres-backed replay, and the in-memory buffer is irrelevant by design.

Closing this thread on those grounds. #1, #3, #4 fixed in c74c53a7.

PR #186, opened 2026-04-29
feat(mcp): freeze global registry at startup, drop per-workspace cohorts

Chen-zexi:
## Summary

- Lift `MCPRegistry` to a process-global frozen snapshot installed once at lifespan startup. After `connect_all` + `freeze`, the registry holds tool schemas in memory and all stdio subprocesses exit; `connect_all`/`disconnect_all` early-return so Sessions can safely share the snapshot.
- Each `Session` now borrows the global instance instead of spawning its own MCP cohort. Per-instance ownership is tracked so `cleanup`/`stop` only tear down what the Session created.
- Backend MCP cost drops from `len(enabled_servers)` per active workspace to **zero idle subprocesses after boot**. The agent's actual tool calls already run via the sandbox-side cohort emitted by `generate_mcp_client_code`, so the backend cohort was read-only after schema discovery.
- Boot-failure hardening: `freeze()` wraps connector teardown in a 10s timeout; `connect_all` drops connectors whose `__aenter__` raised so the snapshot can't carry phantom servers; `set_global_registry` rejects unfrozen instances; lifespan rollback uses `_force_disconnect_all()` so subprocesses are reaped regardless of how far `freeze()` got before raising; the failure log emits only `error_type` (no env-leak risk via raw exception text); `call_tool` raises post-freeze as a defensive guard against future regressions.
- Inline `TODO(option-c)` marks the long-term path: have the sandbox emit its own schemas via a one-shot `--describe` mode at boot, eliminating the backend cohort entirely.

## Test plan

- [x] `uv run pytest tests/unit/` — **3166 passed** (was 3160; +6 new tests)
- [x] `uv run ruff check` on touched files — clean
- [x] New tests cover: `freeze()` terminates subprocesses while keeping tools, `freeze()` idempotency, `connect_all`/`disconnect_all` no-op when frozen, `_force_disconnect_all` runs through frozen state, `call_tool` raises post-freeze, `set_global_registry` rejects unfrozen, `connect_all` drops failed connectors, global registry get/set/clear lifecycle.
- [x] Session borrow/own unit tests cover: borrows global when installed, creates own when no global, `cleanup` skips `disconnect_all` for borrowed global, `cleanup` does tear down owned registry.
- [x] Autouse fixture moved to `tests/unit/core/conftest.py` so `_GLOBAL_REGISTRY` cannot leak across tests outside the registry test class.

## Manual verification

After merge + redeploy, expect:
- Backend startup logs: one `MCP registry frozen servers=10 tools=69` line.
- `pgrep -af 'mcp_servers/' | wc -l` inside the backend container → 0 after boot (was N_workspaces × 10).
- Per-workspace startup latency drops the MCP-connect contribution to ~0ms; lazy log shows `borrowed_global=True`.

### DIFF
--- a/src/ptc_agent/core/mcp_registry.py
+++ b/src/ptc_agent/core/mcp_registry.py
@@ -23,7 +23,7 @@ _devnull = open(os.devnull, "w")  # noqa: SIM115
 
 
 class MCPToolInfo:
-    """Information about an MCP tool."""
+    """Snapshot of a single tool's schema as reported by its MCP server."""
 
     def __init__(
         self,
@@ -32,25 +32,13 @@ class MCPToolInfo:
         input_schema: dict[str, Any],
         server_name: str,
     ) -> None:
-        """Initialize tool info.
-
-        Args:
-            name: Tool name
-            description: Tool description
-            input_schema: JSON schema for tool input
-            server_name: Name of the MCP server providing this tool
-        """
         self.name = name
         self.description = description
         self.input_schema = input_schema
         self.server_name = server_name
 
     def get_parameters(self) -> dict[str, Any]:
-        """Extract parameter information from input schema.
-
-        Returns:
-            Dictionary mapping parameter names to their info
-        """
+        """Return ``{param_name: {type, description, required, default}}`` from input_schema."""
         params = {}
 
         if "properties" in self.input_schema:
@@ -129,11 +117,6 @@ class MCPServerConnector:
     """
 
     def __init__(self, config: MCPServerConfig) -> None:
-        """Initialize server connector.
-
-        Args:
-            config: Server configuration
-        """
         self.config = config
         self.session: ClientSession | None = None
         self.tools: list[MCPToolInfo] = []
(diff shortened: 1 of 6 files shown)


Chen-zexi:
Verified — not actionable as written.

`src/server/app/setup.py` uses **stdlib `logging`** (line 26 `import logging`, line 47 `logger = logging.getLogger(__name__)`), not structlog. `MCPRegistry.freeze()` lives in `src/ptc_agent/core/mcp_registry.py` which **does** use structlog (line 9 `import structlog`). Different logger types, different APIs.

Calling `logging.Logger.warning(msg, key=value)` raises `TypeError: Logger._log() got an unexpected keyword argument 'error'`. We hit exactly this crash on `backend-1` startup yesterday and the printf-style positional args here are the deliberate fix:

```
backend-1 | 2026-04-29 01:05:17,729 - src.server.app.setup - WARNING - Failed to initialize PTC Agent: Logger._log() got an unexpected keyword argument 'error'
```

Applying the suggested diff would re-break lifespan startup with the same error.

@claude please investigate the broader logger inconsistency: `src/server/app/` uses stdlib `logging` while `src/ptc_agent/` uses structlog. Look at whether setup.py (and peers) should migrate to structlog so log aggregators get structured fields uniformly, or whether the split is intentional. If migration is the right call, file a follow-up issue with the scope (which files, what fields/keys to standardize on, any logging.basicConfig vs structlog.configure conflicts to resolve). Do not change setup.py in this PR — it's correct for its current logger.

PR #188, opened 2026-04-29
feat(workflow): canonical terminal status wiring + invariant-pinned reconnect set

Chen-zexi:
## Summary

Closes the gap where `/status` could keep reporting `ACTIVE` for the full Redis TTL window after a workflow died. Every workflow-death path now writes a terminal state to Redis with bounded TTL from one canonical site, and `can_reconnect` reads from a constant pinned by contract tests.

**Production wiring (4 files, +161/-4)**

- **`workflow_tracker.py`** — adds `WorkflowStatus.FAILED` + `SOFT_INTERRUPTED`, the matching `mark_failed` / `mark_soft_interrupted` methods (bounded TTL via `get_redis_ttl_workflow_status`, mirroring `mark_completed` / `mark_cancelled`), and the `TERMINAL_STATUSES` + `RECONNECTABLE_STATUSES` frozensets.
- **`background_task_manager.py`** — wires `tracker.mark_failed` into `_mark_failed`, `tracker.mark_soft_interrupted` into `_mark_soft_interrupted`, and `tracker.mark_cancelled` into `_mark_cancelled` as the canonical cancel site (cancel endpoint, stale-cancel reaper, soft-interrupt-abort all go through it now). Hoists the `WorkflowTracker` import to module level.
- **`handlers/chat/_common.py`** — wires `tracker.mark_failed` into both branches of `handle_workflow_error` (max-retries-exceeded + non-recoverable). Fixes the setup-error path that ran outside `BackgroundTaskManager` and previously left status at `ACTIVE`.
- **`handlers/workflow_handler.py`** — `can_reconnect` now reads `RECONNECTABLE_STATUSES` instead of an inline `[ACTIVE, DISCONNECTED]` list.

**Before:** setup-error and non-recoverable workflows left the Redis status at `ACTIVE` for the full TTL window after the workflow died, and the frontend kept attempting to reconnect to a stream that would never produce events.

## Test Coverage

49 new/updated unit tests across 3 files:

- `test_workflow_tracker.py` — unit tests for `mark_failed` / `mark_soft_interrupted` (TTL, status, metadata, disabled-mode), plus a new `TestStatusSetInvariants` class pinning the contract: `TERMINAL_STATUSES` ⊥ `RECONNECTABLE_STATUSES`, every `WorkflowStatus` is partitioned, and `get_workflow_status` actually reads from `RECONNECTABLE_STATUSES`. Adding a new status without categorizing it now fails CI.
- `test_background_task_terminal_cleanup.py` — extended the three mark-releases tests to assert `tracker.mark_failed` / `mark_cancelled` / `mark_soft_interrupted` are invoked with the expected args. Catches future refactors that drop the wiring.
- `test_handle_workflow_error.py` (new) — covers the max-retries branch, the non-recoverable branch, and tracker-failure resilience (a Redis outage during `mark_failed` must not break the SSE error event).

## Pre-Landing Review

Cleared via `/review` before push. Critical pass clean (no SQL, race, LLM trust, or shell injection); two informational findings from the original commit (TERMINAL_STATUSES was unused; wiring sites had no tests) were both addressed by adding `RECONNECTABLE_STATUSES` + the contract/wiring tests above.

## Verification

- `uv run pytest` (49 touched-file tests + broader `tests/unit/server/`): pass
- `uv run ruff check`: clean
- For end-to-end smoke against real Redis: trigger each terminal path (cancel-endpoint, ESC soft-interrupt, AttributeError setup-error, max-retries) and confirm `redis-cli GET workflow:status:<thread_id>` reports the expected terminal state with `TTL ≈ 3600`.

## Test plan

- [x] `mark_failed` / `mark_soft_interrupted` write the right Redis payload + bounded TTL
- [x] `BTM._mark_failed` / `_mark_cancelled` / `_mark_soft_interrupted` invoke the tracker at the canonical site
- [x] `handle_workflow_error` invokes `tracker.mark_failed` in both branches
- [x] `TERMINAL_STATUSES` ⊥ `RECONNECTABLE_STATUSES` invariant pinned
- [x] `workflow_handler.get_workflow_status` reads from `RECONNECTABLE_STATUSES`
- [ ] (manual smoke) cancel-endpoint flow leaves Redis at `cancelled` with bounded TTL
- [ ] (manual smoke) frontend stops attempting reconnect after a terminal status

### DIFF
--- a/src/server/handlers/chat/_common.py
+++ b/src/server/handlers/chat/_common.py
@@ -889,9 +889,13 @@ async def handle_workflow_error(
                 f"thread_id={thread_id}: {type(e).__name__}: {str(e)[:100]}"
             )
 
+            error_msg = (
+                f"Max retries exceeded ({retry_count}/{MAX_RETRIES}): "
+                f"{type(e).__name__}: {str(e)}"
+            )
+
             if persistence_service:
                 try:
-                    error_msg = f"Max retries exceeded ({retry_count}/{MAX_RETRIES}): {type(e).__name__}: {str(e)}"
                     await persistence_service.persist_error(
                         error_message=error_msg,
                         errors=[error_msg],
@@ -905,6 +909,18 @@ async def handle_workflow_error(
                     logger.error(
                         f"[{log_prefix}] Failed to persist error: {persist_error}"
                     )
+
+            # Push terminal status to Redis so /status reports FAILED with
+            # bounded TTL instead of leaving the key as ACTIVE/DISCONNECTED.
+            # The setup-error path runs outside BackgroundTaskManager's
+            # _mark_failed, so this is the only chance to update tracker.
+            try:
+                await tracker.mark_failed(thread_id, error=error_msg)
+            except Exception as tracker_err:
+                logger.warning(
+                    f"[{log_prefix}] tracker.mark_failed failed for "
+                    f"{thread_id}: {tracker_err}"
+                )
 
             error_data = {
                 "message": f"Workflow failed after {MAX_RETRIES} retry attempts",
@@ -951,6 +967,20 @@ async def handle_workflow_error(
                 )
             except Exception as persist_error:
                 logger.error(f"[{log_prefix}] Failed to persist error: {persist_error}")
+
+        # Mirror the recoverable max-retries branch: push FAILED to Redis with
+        # bounded TTL. Without this, /status keeps reporting ACTIVE for the
+        # full mark_active TTL window after a non-recoverable workflow error.
+        try:
+            tracker = WorkflowTracker.get_instance()
+            await tracker.mark_failed(
+                thread_id, error=f"{type(e).__name__}: {str(e)}"
+            )
+        except Exception as tracker_err:
+            logger.warning(
+                f"[{log_prefix}] tracker.mark_failed failed for "
+                f"{thread_id}: {tracker_err}"
+            )
 
         if handler:
             error_event = handler._format_sse_event(
--- a/src/server/handlers/workflow_handler.py
+++ b/src/server/handlers/workflow_handler.py
@@ -168,7 +168,11 @@ async def get_workflow_status(thread_id: str) -> dict:
         Dict with current status, reconnectability, and progress info
     """
     try:
-        from src.server.services.workflow_tracker import WorkflowTracker, WorkflowStatus
+        from src.server.services.workflow_tracker import (
+            RECONNECTABLE_STATUSES,
+            WorkflowStatus,
+            WorkflowTracker,
+        )
 
         tracker = WorkflowTracker.get_instance()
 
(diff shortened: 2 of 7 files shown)


Chen-zexi:
Verified. Three of the four landed in `45bb742d`:

**1. Permissive metadata assertion** — tightened to `assert "metadata" not in persisted`. The implementation in `_update_status_with_metadata` only writes the key when `metadata` is truthy, so the strict form matches reality and a regression to `{"metadata": {}}` would now fail.

**2. Hardcoded TTL** — both `test_mark_failed_sets_ttl_and_status` and `test_mark_soft_interrupted_sets_ttl` (plus the older `test_mark_completed_sets_ttl`) now read from `get_redis_ttl_workflow_status()`. Config drift no longer silently passes.

**3. `inspect.getsource` brittleness** — replaced with two parametrized behavioral tests:
- `test_get_workflow_status_reconnectable[active|disconnected]` — mocks tracker + BTM + checkpoint, asserts `can_reconnect=True`
- `test_get_workflow_status_terminal_blocks_reconnect[cancelled|completed|failed|soft_interrupted]` — same harness, asserts `can_reconnect=False`

These exercise the actual production decision (membership against `RECONNECTABLE_STATUSES`) instead of pinning a string literal, so a rename or refactor into a helper survives.

**4. Dual `mark_cancelled` on the explicit-cancel path** — intentional and already documented at `background_task_manager.py:2173-2177`. The order is `tracker.mark_cancelled` (in `_handle_sse_disconnect`) → `persistence_service.persist_cancelled` → `manager.cancel_workflow` → BTM `_mark_cancelled` (which calls `tracker.mark_cancelled` again). Removing the explicit call would leave Redis at `active` during the persistence window between disconnect detection and BTM cleanup, so any /status check in that window would lie. Redis SET is idempotent, the second call is cheap, and the BTM site is the canonical one for the four other cancel paths (cancel endpoint, stale-cancel reaper, soft-interrupt-abort, etc.). Keeping both.

54/54 tests green, ruff clean.

Chen-zexi:
Both addressed in `98e11f63`:

**1. Positional vs keyword `error` arg** — `background_task_manager._mark_failed` now calls `tracker.mark_failed(thread_id, error=error)`, matching the keyword form already used in both `chat/_common.py` call sites. Updated the matching `assert_awaited_once_with` in `test_mark_failed_releases_and_signals` to keep contract pinning aligned.

**2. Redundant `AsyncMock` assignments** — `_patch_tracker` no longer assigns `mark_failed`/`mark_cancelled`/`mark_soft_interrupted` individually. `AsyncMock()` auto-specs async attributes on access, so `mock_tracker.mark_failed` resolves to a child `AsyncMock` either way. Calls and `assert_awaited_*` still work, and the doc-comment now explains the auto-spec behavior so a future reader doesn't reintroduce the boilerplate.

54/54 tests green, ruff clean. Ready to merge from my side.

PR #192, opened 2026-05-02
Decouple SSE streaming from workflow via Redis Streams

Chen-zexi:
**Status: DRAFT — not ready for review.** Parking the staged Streams refactor plus the subagent-card-stays-running fix on a single branch so they soak together. The first three commits are the staged refactor (per `~/.claude/plans/checkout-a-branch-off-greedy-badger.md`); the fourth is the bug fix that motivated the deeper review pass.

## Summary

Four bisectable commits:

1. **`refactor(sse): dual-write workflow + subagent events to Redis Streams`** — additive producer-only. Extends `pipelined_event_buffer` with a `stream_key` parameter; both call sites (main workflow buffer, subagent spill) dual-write to `workflow:stream:{thread_id}` / `subagent:stream:{thread_id}:{task_id}` inside the existing MULTI/EXEC. Includes the dirty-resume guard so a fresh-handler XADD with id=1-0 doesn't collide with stale Stream/List/seq state from a prior crashed turn.
2. **`feat(sse): redis-streams consumers behind USE_REDIS_STREAM_SSE flag`** — flag-gated read path. Single `_stream_from_redis_log` XREAD BLOCK loop drives both `stream_from_log` (main) and `stream_subagent_from_log` (per-subagent). Three connection shapes covered: `$` first-connect, `0` replay, `<seq>-0` resume. `threads.py` accepts `Last-Event-ID` header (SSE-spec) in addition to the existing `?last_event_id=` query param. Default off; flip per request.
3. **`test(sse): integration test covering unified consumer paths`** — single integration test exercising replay-from-zero, resume-after-seq, second-tab live-tail, and late-subscriber-after-DEL. Each scenario calls the same helper with a different cursor — that's the architectural win.
4. **`feat(subagent): per-token streaming + stream-end sentinel`** — fixes the "subagent card stays Running" UX. Subagents now stream per-token chunks through `subagent.astream(stream_mode=["values", "messages"])` via a new `_arun_subagent_streaming` wrapper, and write a `subagent_stream_end` sentinel record to the per-task Stream when the astream loop exits. The per-task SSE consumer recognises the sentinel and closes immediately instead of waiting on a 4-8s XREAD BLOCK round-trip plus the post-turn collector to flip `task.completed`. Folds in: single-pass `_classify_subagent_payload` (one `json.loads` max per entry), `active_connections` hooks on the workflow consumer (so abandoned-task cleanup doesn't force-cancel SSE-attached workflows), `_xread_block_ms` floor below `socket_timeout`, and frontend React.memo invalidation for per-token deltas.

## Diff Size
- Total: 24 files changed, 2729 insertions(+), 172 deletions(-)
- Net (excl. tests): 13 files changed, 959 insertions(+), 155 deletions(-)

## Test Coverage

Unit: 3277 passed, 0 failed (full sweep on touched dirs + sibling areas). New unit tests:
- `tests/unit/utils/cache/test_pipelined_event_buffer.py` — dirty-resume guard parity (Stream + List + meta seq), stream_event override
- `tests/unit/server/handlers/chat/test_stream_from_log.py` — `_classify_subagent_payload` 4-kind contract, `stream_from_log` connection-counter hooks, sentinel exit, `_xread_block_ms` floor
- `tests/unit/middleware/test_subagent_token_forwarder.py` (new) — token forwarder reasoning lifecycle (start/complete signals), text/reasoning transitions, finalize sentinel
- `tests/unit/middleware/test_subagent_event_capture.py` — additions covering the moved-token-capture branch
- `tests/unit/middleware/test_subagent_tracer_isolation.py` — extensions

Integration: `tests/integration/test_sse_streams.py` covers the four-scenario unified-consumer path (requires real Redis via `make setup-db`).

## Pre-Landing Review

Already ran `/review` (3 critical-tier verified findings + 4 informational, all addressed). Then ran `/audit-prose` for prose-only cleanup (9 auto-fixes + 1 reviewed apply). All findings closed. Lint clean (`ruff check`) on every touched file.

## Out of scope (subsequent PRs, gated on prod soak)

- **Cutover:** Delete `multiplex_streams`, `live_queues`, `result_buffer`, `captured_events_tail`, `EventCounter`, `StreamEventAccumulator`, `get_buffered_events_redis`, `iter_subagent_events_full`, `drain_task_captured_events`. Replace with `XRANGE - +`-based readers.
- **Drop List spill:** Remove the RPUSH/LTRIM half of `pipelined_event_buffer` once cutover has soaked +24h.
- **Drop shield:** Remove `asyncio.shield(inner_task)` since SSE consumers no longer hold workflow refs.

## Test plan

- [ ] Soak with `USE_REDIS_STREAM_SSE=false` (default) in staging, confirm `XLEN workflow:stream:* == LLEN workflow:events:*` parity for >=24h
- [ ] Flip `USE_REDIS_STREAM_SSE=true` in staging, browser-test that subagent cards flip to "Completed" promptly when subagent finishes streaming (the original sentinel fix)
- [ ] Browser-test reconnect scenarios: kill mid-stream, reopen with `Last-Event-ID`, verify resume
- [ ] Browser-test second-tab on the same thread: both tabs receive identical event stream
- [ ] Confirm abandoned-task cleanup at 6h doesn't force-cancel long-running SSE-attached workflows

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -144,6 +144,14 @@ REDIS_PASSWORD=redis
 # Override config.yaml redis.max_connections (default 150). Range [1, 10000].
 # REDIS_MAX_CONNECTIONS=300
 
+# =============================================================================
+# SSE — Redis Streams consumer (in-progress refactor)
+# =============================================================================
+# When true, both first-connect and reconnect SSE flow through the XREAD-BLOCK
+# consumer reading workflow:stream:* / subagent:stream:* keys. The dual-write
+# producer is always on; this flag only switches the read side. Default: false.
+# USE_REDIS_STREAM_SSE=false
+
 # Postgres connection pool overrides. Ranges [1, 500].
 # POSTGRES_CHECKPOINTER_POOL_MAX=25
 # POSTGRES_CONVERSATION_POOL_MAX=50
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -401,6 +401,15 @@ def is_subagent_event_redis_spill_enabled() -> bool:
     return get_infrastructure_config().background_execution.spill_subagent_events_to_redis
 
 
+def is_use_redis_stream_sse_enabled() -> bool:
+    """Return True when ``USE_REDIS_STREAM_SSE`` env var is set.
+
+    Routes both first-connect and reconnect SSE through the XREAD BLOCK
+    consumer (``workflow:stream:*`` / ``subagent:stream:*``). Default: False.
+    """
+    return os.getenv("USE_REDIS_STREAM_SSE", "false").lower() in ("true", "1", "yes")
+
+
 def get_sse_drain_timeout() -> float:
     """Get seconds to wait for per-task SSE drain before clearing events."""
     return get_infrastructure_config().background_execution.sse_drain_timeout
--- a/src/ptc_agent/agent/middleware/background_subagent/event_capture.py
+++ b/src/ptc_agent/agent/middleware/background_subagent/event_capture.py
@@ -169,80 +169,14 @@ class SubagentEventCaptureMiddleware(AgentMiddleware):
             try:
                 ai_msg = response.result[0] if response.result else None
                 if ai_msg:
-                    from src.llms.content_utils import format_llm_content
-
-                    formatted = format_llm_content(ai_msg.content)
                     tool_calls = getattr(ai_msg, "tool_calls", None) or []
                     agent_id = self._get_agent_id(tool_call_id)
                     msg_id = getattr(ai_msg, "id", f"msg-{tool_call_id}")
 
-                    # Emit reasoning with start/complete signals so frontend
-                    # can initialize currentReasoningIdRef properly
-                    if formatted.get("reasoning"):
-                        await self.registry.append_captured_event(
-                            tool_call_id,
-                            {
-                                "event": "message_chunk",
-                                "data": {
-                                    "agent": agent_id,
-                                    "id": msg_id,
-                                    "role": "assistant",
-                                    "content": "start",
-                                    "content_type": "reasoning_signal",
-                                },
-                                "ts": time.time(),
-                            },
-                        )
-                        await self.registry.append_captured_event(
-                            tool_call_id,
-                            {
-                                "event": "message_chunk",
-                                "data": {
-                                    "agent": agent_id,
-                                    "id": msg_id,
-                                    "role": "assistant",
-                                    "content": _truncate_content(formatted["reasoning"]),
-                                    "content_type": "reasoning",
-                                    "finish_reason": None,
-                                },
-                                "ts": time.time(),
-                            },
-                        )
-                        await self.registry.append_captured_event(
-                            tool_call_id,
-                            {
-                                "event": "message_chunk",
-                                "data": {
-                                    "agent": agent_id,
-                                    "id": msg_id,
-                                    "role": "assistant",
-                                    "content": "complete",
-                                    "content_type": "reasoning_signal",
-                                },
-                                "ts": time.time(),
-                            },
-                        )
-
-                    # Emit text chunk if present
-                    if formatted.get("text"):
-                        await self.registry.append_captured_event(
-                            tool_call_id,
-                            {
-                                "event": "message_chunk",
-                                "data": {
-                                    "agent": agent_id,
-                                    "id": msg_id,
-                                    "role": "assistant",
-                                    "content": _truncate_content(formatted["text"]),
-                                    "content_type": "text",
-                                    "finish_reason": "tool_calls"
-                                    if tool_calls
-                                    else "stop",
-                                },
-                                "ts": time.time(),
-                            },
-                        )
-
+                    # Text/reasoning tokens are forwarded per-token by
+                    # ``_SubagentTokenForwarder``; capturing them here too
+                    # would duplicate every chunk. Tool calls have no streaming
+                    # counterpart so they stay captured here.
                     if tool_calls:
                         await self.registry.append_captured_event(
                             tool_call_id,
--- a/src/ptc_agent/agent/middleware/background_subagent/middleware.py
+++ b/src/ptc_agent/agent/middleware/background_subagent/middleware.py
@@ -270,6 +270,9 @@ class BackgroundSubagentMiddleware(AgentMiddleware):
                     await cache.delete(
                         f"subagent:events:meta:{self.registry.thread_id}:{task.task_id}"
                     )
+                    await cache.delete(
+                        f"subagent:stream:{self.registry.thread_id}:{task.task_id}"
+                    )
             except Exception:
                 logger.warning(
                     "Failed to clear Redis spool on resume; replay may include stale events",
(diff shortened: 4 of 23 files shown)


Chen-zexi:
Thanks — all four addressed in 612daa46.

**1. Cursor stall (blocking).** Fixed. `cursor = entry_id` now runs unconditionally at the top of the inner loop, before either `continue`. New regression test `test_cursor_advances_past_skipped_last_entry` locks the contract: round-1 batch ends with a `{}` (missing `b"event"`) entry, and round-2's XREAD must use that entry's id as the cursor, not the prior successfully-yielded one. Without the fix the test asserts `cursors[1] == b"6-0"` would fail with `b"5-0"` and the consumer would loop forever.

**2. `is` -> `==`.** Fixed in `_dispatch`. The interning comment was correct as a CPython implementation detail but the bot is right that it's not a language guarantee — switched to `==` and updated the comment so future readers don't reach for `is` again.

**3. `on_detach` timing on sentinel exit.** Fixed. Both branches of `stream_subagent_from_log` (with-task and no-task) now manage the inner generator explicitly with `try / finally / inner.aclose()`. The inner generator's `finally` block (which decrements `sse_consumer_count` and sets `sse_drain_complete`) now fires synchronously the moment we `return` on a sentinel, not whenever FastAPI gets around to GCing the outer generator. This closes the race against tight `sse_drain_complete` waits in `_await_drain_and_cleanup_tasks`.

**4. `wait_for` timeout window.** Documented inline. The post-dispatch race is real (sub-500-ms but real) — added a comment at the lock-release boundary explaining the failure mode and noting that `terminal_check` is the safety net when the sentinel is lost.

Lint clean, 247 unit tests pass on touched dirs.

Chen-zexi:
Verified all three findings before acting; folded into the fix commit (now `c094668a`, force-pushed).

**1. Redundant `None`-guard.** ✅ Verified at `stream_reconnect.py:71-73`. Dropped — `_stream_from_redis_log` maps both `None` and `<= 0` to cursor `0`, so the ternary was a no-op. Now passes `last_event_id` through directly. Comment updated to document the cursor mapping.

**2. Fast-path comment.** ✅ Verified the assumption (`registry._spill_record_to_redis` always emits `id:` first). Added an inline comment at `_classify_subagent_payload`'s fast path naming the producer site and explaining what happens if the format widens — future widening will fall through to the JSON-decode path and classify as `_PAYLOAD_UNKNOWN` (still passes through correctly, just bypasses the intended fast path). Audit-both-sites note included.

**3. "Unreachable `return`".** ❌ Refuted on the literal claim — the `return` was reachable AND necessary. The "with-task" code immediately below `if task is None:` is the implicit else and dereferences `task`; without the `return`, exhausting the no-task inner loop would crash with `AttributeError` on `task.completed` in the with-task `terminal_check`. ✅ Underlying readability concern was real, addressed via a refactor: `if task is None: ... else: ...` so both branches set up `inner` and share a single `try/finally`. The bare `return` goes away, the structure is self-documenting, behavior is identical.

Lint clean, 276 unit tests pass on touched dirs (5 new tests across the two review rounds — `test_cursor_advances_past_skipped_last_entry`, `test_classify_subagent_payload_distinguishes_all_four_kinds`, `test_stream_from_log_bumps_workflow_connection_counter`, plus 2 dirty-resume guard tests).

Branch is back to a clean 5-commit structure: `c4109399 refactor(sse): dual-write` → `a9a23dcb feat(sse): consumers behind flag` → `8ce04aca test(sse): integration` → `0534b28d feat(subagent): per-token + sentinel` → `c094668a fix(sse): review-derived fixes`.

Chen-zexi:
Both verified and folded into `d840f030`.

**1. `terminal_seen` reset comment.** ✅ Verified. Comment said "whenever we emit new events" but the reset is unconditional after any non-empty XREAD batch. Producer always writes `b"event"` so all-skipped batches don't occur in practice, but the comment was misleading. Rewrote to match the actual reset condition and added an inline note that the all-skipped case relies on a producer invariant — if that invariant changes, revisit.

**2. Three chained `try/except` import guards.** ✅ Verified concern. Collapsed into a single `try/except Exception: return` covering the two settings/cache imports and the flag check. Added a one-line comment explaining the defensive intent (broken-deployment guard, not a normal failure mode). Same behavior, cleaner read.

Lint clean, 248 unit tests pass on touched dirs. Force-pushed (`c094668a → d840f030`); 5-commit structure preserved.

Chen-zexi:
Verified both, folded into `aa818556`.

**1. Shared logger from `_common`.** ⚠️ Bot's premise is partly wrong: `_common.py:13` is stdlib `logging`, not structlog — the imported `logger` is just `logging.getLogger("src.server.handlers.chat_handler")`, no structlog request-context processors. So there's no structured-context benefit. **However**, the consistency point is valid: every other file in `src/server/handlers/chat/` (`flash_workflow`, `llm_config`, `ptc_workflow`, `steering`, `stream_reconnect`) imports `logger` from `._common`. Switched for stylistic consistency — all chat-handler logs now route under the same logger name.

**2. `asyncio.TimeoutError` handshake participation.** ✅ Verified. A wedged XREAD setting `terminal_seen = True` means the next empty BLOCK round triggers exit (not literally "two empty rounds"). Comment now spells out the actual dwell time: `(block_ms + 2s) + block_ms` before exit, long enough to drain a trailing tail. Pure documentation — behavior unchanged.

Lint clean, 27 chat-handler unit tests pass. Force-pushed (`d840f030 → aa818556`).

Chen-zexi:
Verified all three nits — no code changes warranted.

**1. Invalid `task_id` 38 s stall.** Timing is correct (30 s startup wait + 8 s two-empty-round handshake at default `block_ms=4 s`). The 30 s window exists specifically to tolerate the registry-registration race in normal subagent startup; trimming it would make legit late-attaches miss events. Adding a `subagent_error` SSE would also be a no-op today — `useChatMessages.ts` doesn't handle a `subagent_error` event type, and the per-task stream's completion semantics are signalled by the response closing (not a terminal event). Wiring up a frontend handler + adding an error path for an in-practice-never-happens case crosses into product UX territory; deferring.

**2. Dirty-resume guard comment + `stream_event` arg.** The function-level docstring at `redis_cache.py:620-626` already documents the `stream_event` override and explains the main vs subagent payload split:

> ``stream_event`` defaults to ``event`` but lets callers store a different wire format in the Stream than in the List — used by the subagent producer, which keeps JSON records in the legacy List for in-flight reads but writes pre-rendered SSE wire strings to the Stream so the consumer doesn't need a JSON-render step.

The inline guard comment focuses on the three reset hazards (stream id monotonicity, `LLEN == XLEN` parity, `seq` drift) — orthogonal concern from the payload-format split. Repeating the docstring inline would just be duplication.

**3. `_classify_subagent_payload` empty-string semantics.** Comment itself says "no change needed — just noting." Producer (`registry._spill_record_to_redis`) always writes either an SSE wire string or a JSON record; empty payloads are not on the writer contract. The fast-path comment (`stream_from_log.py:316-323`) already flags the producer-format dependency for future audits.

PR #193, opened 2026-05-03
fix(build): default PROVIDER=docker so 'make up' builds sandbox image

mpesce:
## Summary
- Default `PROVIDER` to `docker` in the Makefile so plain `make up` triggers `_sandbox-prepare`.
- Skip the sandbox image rebuild when `langalpha-sandbox:latest` already exists (~25 min saved per `make up`).
- No behavior change for `make up PROVIDER=daytona` — Daytona users still opt out of the local image build.

## Bug
Plain `make up` (no `PROVIDER` argument) silently skips `_sandbox-prepare` because `ifeq ($(PROVIDER),docker)` is false when `PROVIDER` is unset. The runtime, however, defaults to the docker sandbox provider via `SANDBOX_PROVIDER` in `.env`, so the first workspace creation 500s with:

> Docker image 'langalpha-sandbox:latest' not found and Dockerfile.sandbox not found in any search path

`Dockerfile.sandbox` isn't bind-mounted into the backend container, so the runtime can't lazy-build it either — the host has to build it ahead of time. README and the `make up` help text both imply `make up` is sufficient on its own.

## Test plan
- [ ] `docker rmi langalpha-sandbox:latest && make up` — should build the sandbox image then start the stack
- [ ] `make up` (image already present) — should print "skipping build" and start the stack quickly
- [ ] `make up PROVIDER=daytona` — should NOT build the sandbox image
- [ ] First workspace creation succeeds (no 500)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,11 @@ config: ## Interactive setup wizard (LLM, data, sandbox, search)
 # ---------------------------------------------------------------------------
 # Docker Compose (full-stack)
 # ---------------------------------------------------------------------------
-up: _sandbox-prepare ## Start full stack (PROVIDER=docker|daytona)
+# Default to the docker sandbox provider so plain `make up` works out of the
+# box. Override with `make up PROVIDER=daytona` to skip the sandbox image build.
+PROVIDER ?= docker
+
+up: _sandbox-prepare ## Start full stack (PROVIDER=docker|daytona, default docker)
 	SANDBOX_PROVIDER=$(PROVIDER) docker compose up --build
 
 down: ## Stop all Docker Compose services
@@ -19,11 +23,16 @@ clean: down ## Stop everything and remove stale sandbox containers
 	@docker rm -f $$(docker ps -aq --filter "name=langalpha-sandbox") 2>/dev/null || true
 	@echo "Clean."
 
-# Build sandbox image only when provider needs it (docker)
+# Build sandbox image only when provider needs it (docker), and only if the
+# image isn't already present locally — avoids a ~25 min rebuild on every `up`.
 _sandbox-prepare:
 ifeq ($(PROVIDER),docker)
-	@echo "Building sandbox image for docker provider..."
-	docker build -f Dockerfile.sandbox -t langalpha-sandbox:latest .
+	@if ! docker image inspect langalpha-sandbox:latest >/dev/null 2>&1; then \
+		echo "Building sandbox image for docker provider..."; \
+		docker build -f Dockerfile.sandbox -t langalpha-sandbox:latest .; \
+	else \
+		echo "Sandbox image langalpha-sandbox:latest already present, skipping build."; \
+	fi
 endif
 
 # ---------------------------------------------------------------------------


Makefile
```diff
@@ -8,7 +8,11 @@ config: ## Interactive setup wizard (LLM, data, sandbox, search)
 # ---------------------------------------------------------------------------
 # Docker Compose (full-stack)
 # ---------------------------------------------------------------------------
-up: _sandbox-prepare ## Start full stack (PROVIDER=docker|daytona)
+# Default to the docker sandbox provider so plain `make up` works out of the
+# box. Override with `make up PROVIDER=daytona` to skip the sandbox image build.
+PROVIDER ?= docker
```
Chen-zexi, line 13: Could this read from `.env` first? `make config` already writes `SANDBOX_PROVIDER` there, so users who picked Daytona during setup will still trigger a Docker build under this default — and the `SANDBOX_PROVIDER=$(PROVIDER)` line below silently overrides their `.env` value too.

```suggestion
# PROVIDER resolution: command line > .env > docker fallback.
ENV_PROVIDER := $(shell [ -f .env ] && sed -nE 's/^SANDBOX_PROVIDER=(.*)$$/\1/p' .env | tail -n1 | xargs)
PROVIDER     ?= $(if $(ENV_PROVIDER),$(ENV_PROVIDER),docker)
```

Makefile
```diff
@@ -19,11 +23,16 @@ clean: down ## Stop everything and remove stale sandbox containers
...
 ifeq ($(PROVIDER),docker)
-	@echo "Building sandbox image for docker provider..."
-	docker build -f Dockerfile.sandbox -t langalpha-sandbox:latest .
+	@if ! docker image inspect langalpha-sandbox:latest >/dev/null 2>&1; then \
+		echo "Building sandbox image for docker provider..."; \
+		docker build -f Dockerfile.sandbox -t langalpha-sandbox:latest .; \
+	else \
+		echo "Sandbox image langalpha-sandbox:latest already present, skipping build."; \
+	fi
 endif
```
Chen-zexi, line 36: We already publish `ghcr.io/ginlix-ai/langalpha/sandbox:latest` from `.github/workflows/sandbox-image.yml`, and `sandbox-integration.yml` pulls it instead of building. Could `_sandbox-prepare` do the same? Pull is ~30s vs the 25 min cold build, and `:latest` tracks main so the staleness hazard goes away on its own.

```suggestion
# Sandbox image: prefer local edits → cached image → GHCR pull → local build.
SANDBOX_IMAGE := ghcr.io/ginlix-ai/langalpha/sandbox:latest

_sandbox-prepare:
ifeq ($(PROVIDER),docker)
	@if [ -n "$$(git status --porcelain Dockerfile.sandbox 2>/dev/null)" ]; then \
		echo "Dockerfile.sandbox has local changes — building..."; \
		docker build -f Dockerfile.sandbox -t langalpha-sandbox:latest .; \
	elif docker image inspect langalpha-sandbox:latest >/dev/null 2>&1; then \
		echo "Sandbox image already present."; \
	elif docker pull $(SANDBOX_IMAGE) >/dev/null 2>&1; then \
		docker tag $(SANDBOX_IMAGE) langalpha-sandbox:latest; \
	else \
		echo "GHCR pull failed, building locally..."; \
		docker build -f Dockerfile.sandbox -t langalpha-sandbox:latest .; \
	fi
endif
```

A `make rebuild-sandbox` target that does `docker rmi -f langalpha-sandbox:latest && $(MAKE) _sandbox-prepare` would be a nice companion for "give me whatever main is on now."

Chen-zexi (commented):
Thanks for picking this up. Two suggestions inline that I think are worth folding in before this lands.

PR #196, opened 2026-05-04
refactor(sse): cutover to Redis Streams + post-cutover follow-ups

Chen-zexi:
## Summary

Cuts the SSE pipeline over to Redis Streams as the sole authoritative
event log, then layers four follow-ups on top from review and
post-cutover cleanup.

**Cutover.** Every first-connect, reconnect, second tab, and per-subagent
SSE consumer now goes through one XREAD BLOCK loop in `stream_from_log`
/ `stream_subagent_from_log`. The dual-path machinery is gone:
`multiplex_streams` + `_keepalive_task` from the producer, `live_queues`
+ `result_buffer` from `TaskInfo`, `new_event_signal` +
`drain_task_captured_events` from the subagent path,
`stream_live_events` + `_handle_sse_disconnect` from `_common.py`, and
the `USE_REDIS_STREAM_SSE` flag plumbing. Three obsolete test files go
with them. The producer is now a thin
`async for ev in graph.astream(): yield self._format_sse_event(ev)`;
keepalives live in the consumer's BLOCK timeout. The List spill stays
live so a revert still has full data behind the List-replay readers; it
will come out in a follow-up once this has soaked.

**Tool-flow + reconnect fixes.** The streaming handler narrows its
tool-node AI-chunk gate so `ToolMessage` content survives — previously
`web_fetch` and similar tools leaked the inner extraction LLM's
reasoning while dropping the actual tool return. The subagent token
forwarder gains symmetric tuple-index guards in
`_arun_subagent_streaming` so an upstream LangGraph stream-mode
contract change can't raise `IndexError`. `stream_from_log`'s
`terminal_check` now treats a missing `TaskInfo` (`status=None`) as
terminal rather than "still running", so reconnects on
already-cleaned-up turns exit cleanly instead of polling forever.

**Dead-config removal.** `event_storage_fallback_to_memory` only
existed as a safety net for the pre-cutover List-based path. With
Streams authoritative, the fallback never triggers. Strips the
Pydantic field, the settings getter, the yaml key, the
`.env.example` note, and the unused `TaskInfo.event_storage_fallback`
field; tests drop the matching patch decorators.

**History-replay dedup.** History bubbles were keyed on
`` `${role}-${pairIndex}-${Date.now()}` ``, so a second
`loadConversationHistory` invocation (e.g. via the failed-reconnect
`setReloadTrigger` increment) produced fresh ids that the
non-deduping `setMessages` reducer accepted as new bubbles —
duplicating the visible transcript. Switched to deterministic
`history-{role}-{pairIndex}` and added reset-before-replay so re-runs
start from a clean slate. New tests pin both the deterministic-id
contract and the 404-vs-non-404 error contract on the load path.

**Cleanup.** Stale docstrings and comments scrubbed across files
whose behavior is otherwise unchanged: references to deleted
machinery (`multiplex_streams`, `_handle_sse_disconnect`, in-memory
event tail terminology, the gated stream flag), and a rewrite of
`_spill_record_to_redis`'s docstring to accurately describe the dual
write — per-task Redis List for the post-turn collector + per-task
Stream for live SSE consumers.

## Diff Size
- Total: 29 files, +865 / -3066
- Net (excl. tests): 16 files, +288 / -546

## Commits

- `51aa1714` refactor(sse): cutover to Redis Streams; remove List buffer + multiplex_streams + tail deques
- `b60a69d5` fix(sse): post-cutover review fixes for tool flow + reconnect
- `4e5a5a51` chore(config): drop event_storage_fallback dead path; refresh BTM stale comments
- `e0ba0834` fix(web): deterministic history-replay bubble ids
- `cccfec0a` chore(sse): clean up stale docstrings and comments

## Test plan
- [x] `uv run ruff check src/` — clean on touched files (pre-existing
      errors in `llms/`, `workspace_manager.py`, etc. unchanged)
- [x] `cd web && pnpm lint` — 0 errors (pre-existing warnings unchanged)
- [x] ChatAgent vitest suite — passes
- [x] Backend unit tests for touched files — pass
- [x] Manual smoke: open chat, run a query, kill mid-stream, reconnect,
      verify resume + no duplicate bubbles
- [x] Manual smoke: explicit cancel via `/cancel` endpoint while SSE is
      open — verify CANCELLED turn is persisted with token usage


Chen-zexi:
the last claude review seems based on a prior commit, @claude verify the fix and run an independent review.

PR #199, opened 2026-05-06
fix(workspaces): CJK-safe Content-Disposition + charset-detect /files/read

Chen-zexi:
## Summary

Two related fixes for non-ASCII (CJK, emoji, accented) handling on workspace
file endpoints. Both bugs trace to the same root cause class: code path
assumes ASCII / UTF-8 even though the bytes flowing through it can legitimately
be otherwise.

### Bug 1: `/files/download` returns 500 for non-ASCII filenames

`Content-Disposition` headers were built by interpolating the raw filename:

```python
headers={"Content-Disposition": f'attachment; filename="{filename}"'}
```

Starlette encodes header values as latin-1, so any character outside latin-1
(CJK, emoji, "café") makes the encoder raise `UnicodeEncodeError` and the
response 500s. Affects every workspace download for a file with a non-ASCII
name. Memo had already worked around this internally with a private helper.

**Fix:** extract memo's helper into a shared `src/server/utils/http_headers.py`
and apply it to all four `Content-Disposition` sites (`workspace_files.py`,
`public.py` ×2, `memo.py`). Emits the RFC 6266 dual-form header — an ASCII
fallback plus `filename*=UTF-8''<percent-encoded>` so modern clients recover
the original unicode name. Strips `\r\n\"` defensively, and the `disposition`
param is `Literal[\"inline\", \"attachment\"]` so callers can't smuggle a
CRLF-laden value past the filename-only sanitization.

### Bug 2: `/files/read` returns 415 for non-UTF-8 text files

Agent-generated reports in non-UTF-8 locales (mainland China GBK, Taiwan / HK
Big5, Japan Shift-JIS, Korea EUC-KR, Western European single-byte) routinely
land on disk in the system's default codec. The endpoint did UTF-8-only
decode and 415'd the rest with a misleading \"binary file\" message.

**Fix:** UTF-8 fast path stays. On `UnicodeDecodeError`, fall back to
`charset-normalizer.from_bytes(...).best()` — the conventional Python encoding
detector (~70 encodings, chaos-scored, used by `requests` itself). Two
guardrails so genuinely binary content still 415s:

- `chaos < 0.1` — empirically PNG headers score ~0.14, real CJK text scores
  well under 0.05
- `min 8 bytes` — a 3-byte sequence will happily match `cp1006` (Urdu) with
  chaos=0.000; the floor catches that

Promotes `charset-normalizer` from transitive (via `requests`,
`pdfminer-six`) to direct dependency so a future removal can't silently break
this path.

## Diff Size

- Total: 10 files changed, 289 insertions(+), 49 deletions(-)
- Net (excl. tests): 7 files changed, 99 insertions(+), 45 deletions(-)

## Why this shape

- **Root-cause fix, not a band-aid.** Starlette's latin-1 header encoder is
  the underlying constraint; RFC 6266 was written for exactly this case.
- **Comprehensive coverage.** Audited every `Content-Disposition` site in
  `src/` via grep — all four now route through one helper. No drift surface.
- **Conventional CJK handling.** A narrow GB18030 fallback would silently
  mis-decode Big5 / Shift-JIS as gibberish. `charset-normalizer` is the
  conventional library; the chaos threshold + min-bytes floor are the
  guardrails the library lacks by default.

## Test Coverage

19 new tests across two files:

- `tests/unit/server/utils/test_http_headers.py` (9 tests): ASCII pass-through,
  generic CJK, emoji, quote / CR / LF stripping, both dispositions, empty
  fallback, space-only fallback, all-unsafe filename, backslash hardening.
- `tests/unit/server/app/test_workspace_files_decode.py` (10 tests):
  UTF-8 / ASCII pass-through, GBK / Big5 / Shift-JIS decoded, PNG header
  rejected, random binary rejected, short non-UTF-8 rejected, UTF-16
  BOM-only rejected. Self-verifying fixtures (\`_is_non_utf8\`) prevent
  fixture rot — if someone \"fixes\" a fixture to be UTF-8 the test fails
  loudly rather than silently bypassing the fallback path.
- Two existing memo tests rewired to the shared helper.

## Test plan

- [x] \`uv run pytest tests/unit/ -v\` — 3260 / 3260 pass
- [x] \`uv run ruff check\` on touched files — clean
- [ ] Post-merge: round-trip a workspace file with a CJK filename through
      \`/files/download\` and verify the response is 200 with
      \`Content-Disposition: ...; filename*=UTF-8''...\`
- [ ] Post-merge: round-trip a GBK-encoded \`.md\` file through \`/files/read\`
      and verify the body decodes correctly instead of 415-ing


Chen-zexi:
**1. Misleading test name/docstring** — addressed in e601d724.

Renamed `test_all_unsafe_filename_falls_back` → `test_pure_cjk_filename_ascii_token_is_underscores`. Comment rewritten to say what the test actually exercises (CJK chars become `__` via the regex sub; fallback is NOT used here — `test_space_only_falls_back` is the fallback-path case). Also tightened the assertion from `'filename="' in header` (tautologically true) to `'filename="__"' in header` so the underscore-token behavior is now pinned.

**2. `fallback` parameter unvalidated** — declined.

All four call sites pass string literals (`"memo"`, `"download"`) from this codebase. `fallback` is an internal API parameter, not a system boundary that takes user input. Adding `\r\n"` stripping for a scenario that can't happen today would be speculative defense for a hypothetical future caller. If someone later threads user data through `fallback`, hygiene belongs at that new boundary, not pre-applied here.

**3. Sync CPU work in async handler** — declined.

Acknowledged the `from_bytes(...).best()` call blocks the event loop on the non-UTF-8 fallback path. Quoting your own caveat: "the practical impact is low today." UTF-8 fast path covers the common case; `charset-normalizer` samples internally so even typical agent reports (KB to a few hundred KB) stay sub-100ms. If this endpoint starts seeing multi-MB non-UTF-8 files in production, wrapping in `asyncio.to_thread` is a one-line fix at that point. Pre-paying for a hypothetical bottleneck would be premature.

PR #213, opened 2026-05-19
refactor(sse): drop Redis List writes from event-buffer pipeline

Chen-zexi:
## Summary

Follow-up to #196. Drops the now-dead Redis List half of the SSE event-buffer dual-write. The per-thread Redis Stream is the sole durable store for SSE replay + post-turn persistence. The in-memory deque that backed the subagent post-turn collector is gone.

Every SSE consumer was already migrated to read the Stream during the #196 cutover, so the List writes had been dead-letter for a release. This PR strips them, plus the bounded deque (`captured_events_tail`) and its config knob (`in_memory_event_tail_max_events`).

**Two bisectable commits:**

- **Main workflow stream-only** — `pipelined_event_buffer.events_key` becomes optional; main caller passes `None`. Secretary tool's active-thread reader migrates from List to `XREVRANGE workflow:stream:{tid}`. `_buffer_event_redis` now bails with a WARNING when the SSE id can't be parsed instead of advancing the meta `seq` past an event that was never written.
- **Subagent stream-only** — XADD carries dual fields: `b"event"` (pre-rendered SSE wire for live consumers) + `b"record"` (JSON for the post-turn collector). `iter_subagent_events_full` rewritten to `XRANGE` over the per-task stream. Deque + List + tail-size config + `_resolve_tail_maxlen` deleted. `pipelined_event_buffer` becomes purely `meta hash + Stream`.

**Rolling-deploy hygiene.** Terminal + resume cleanup paths sweep the legacy List key for one release so stale RPUSH state from a pre-cutover worker doesn't sit in Redis until the 24h TTL.

## Diff Size

- Total: 17 files changed, 637 insertions(+), 604 deletions(-)
- Net (excl. tests): 6 files changed, 182 insertions(+), 233 deletions(-)

## Verification

- Unit suite: 3320 passed
- Frontend suite: 1074 passed (99 files)
- Lint: clean on touched files
- New unit coverage on all changed branches: dual-payload XADD, XRANGE collector, XREVRANGE secretary read, dirty-resume guard, event-id bail, legacy-key cleanup sweeps
- Manual smoke: cold-start SSE, reconnect mid-stream, subagent dual-payload, multi-turn dirty-resume — all good
- Integration test rewritten to assert single-stream `{event, record}` invariants against real Redis

## Tradeoff

Pre-cutover XADD entries (single `b"event"` field, no `b"record"`) are SKIPPED by the new post-turn collector. Live SSE replay is unaffected. The `subagent_history_truncated` warning fires when this happens. SOP for this release: drain long-running subagents before deploying.

## What's next on this track

This is one in a series of PRs decoupling SSE delivery from in-process workflow state to unlock multi-uvicorn-worker deployment. Remaining work:

- **Track 1 finish:** drop the `asyncio.shield` around the workflow task — relic of the pre-Streams `multiplex_streams` design; no SSE consumer holds a reference to the workflow task anymore.
- **Track 2 — multi-worker decoupling:**
  - Cancel + interrupt via Redis (currently in-memory event flags, lost across workers)
  - Subagent terminal status in Redis (cross-worker SSE GET must see terminal even without registry hit)
  - Main status endpoint Redis-only (drop in-memory `BackgroundTaskManager.tasks` reads)
  - Sticky routing for workflow-owning routes (nginx `hash \$arg_thread_id consistent`)
  - Singleton leader election for the automation scheduler, market-data WS fan-out, and workspace-create lock

After that ships, `uvicorn --workers N` is unlocked.

## Test plan

- [x] \`uv run pytest tests/unit/\`
- [x] \`cd web && pnpm vitest run\`
- [x] \`uv run ruff check src/\`
- [x] Manual SSE end-to-end smoke (live + reconnect + subagent + multi-turn)


Chen-zexi:
both findings verified, addressing Finding 1 in 68b901c6, accepting Finding 2.

**Finding 1 — `event` arg is dead on subagent path:** confirmed. `stream_event` always wins in `raw = stream_event if stream_event is not None else event`, so the subagent caller's `event=payload` was silently discarded.

Fixed in 68b901c6 by making the contract explicit:
- `event: Optional[str] = None`, kwarg-only after `meta_key`/`max_size`/`ttl`
- Main-workflow callers pass `event`; subagent callers pass `stream_event` + `stream_record`
- ValueError fires if a stream write is requested without either payload, so the outer except returns `(False, 0)` instead of silently advancing `meta seq` past an event that never lands
- Dropped `event=payload` from `_spill_record_to_redis`
- New test `test_stream_write_without_payload_returns_failure` covers the invariant
- Drive-by fix: pre-existing dormant bug in `_log_error` where `redis.exceptions` was referenced on the `redis.asyncio` alias — the new test was the first to flow an exception through that path

**Finding 2 — `iter_subagent_events_full` always issues XRANGE now:** accepting. Confirmed: post-cutover, every subagent eats one XRANGE RTT at turn end regardless of event count, where the old deque short-circuited for short runs. Restoring the shortcut would re-introduce the in-memory tail this PR set out to delete, so net negative. The path is off the streaming hot path, runs at turn close only, and is bounded by Stream MAXLEN. Will watch Redis command metrics post-deploy; the natural follow-up (PR M2 onward) doesn't move this needle further.

PR #215, opened 2026-05-19
fix(chat): dedupe HITL card on reconnect after stale-interrupt refresh

Chen-zexi:
## Summary

When a user refreshes the chat page during the rare window where the workflow status is still `ACTIVE` but the persisted `conversation_responses.sse_events` already includes an unresolved `interrupt` event, the question/proposal HITL card renders twice — once attached to the history bubble (populated by `replayThreadHistory`'s interrupt handler) and once on the reconnect bubble (populated by the live SSE interrupt handler after reconnect re-delivers the same event from the Redis stream).

The underlying cause is a cursor-format mismatch: `replayThreadHistory` emits local `id: <seq>` integers (small, e.g. 47) which set the frontend's `lastEventIdRef`. The reconnect endpoint forwards this to Redis as the XREAD cursor, but Redis stream entries use ms-timestamp IDs (~`1701234567890`) — all vastly larger than any replay seq, so the cursor effectively replays the entire Redis log including the buffered `interrupt` event.

The fix runs entirely inside `reconnectToStream`: before appending the new reconnect bubble, walk `unresolvedHistoryInterruptRef.current` to strip the history-replayed interrupt segments (`user_question`, `create_workspace`, `start_question`, `ptc_agent`, `delete_workspace`, `stop_workspace`, `delete_thread`, `plan_approval`) and their backing entity-map entries from the history bubble, then redirect each `unresolvedHistoryInterruptRef[].assistantMessageId` to the new `assistant-reconnect-${ts}` id so the `tool_call_result` history-resolver writes resolution status to the bubble where the proposal now actually lives. Non-interrupt content on the history bubble (text, reasoning, tool calls) is preserved; the existing empty-bubble guard cleans up bubbles that contained only interrupt content. The logic is gated on `unresolvedHistoryInterruptRef.current.length > 0`, so it's a no-op in the common case.

## Test plan

- [ ] Manually reproduce by triggering an unanswered interrupt and refreshing during the `ACTIVE` -> `INTERRUPTED` window — without this fix, the card duplicates; with it, only one card renders.
- [ ] All 75 ChatAgent hook unit tests pass; lint and tsc clean on the modified file.

### DIFF
--- a/web/src/pages/ChatAgent/hooks/useChatMessages.ts
+++ b/web/src/pages/ChatAgent/hooks/useChatMessages.ts
@@ -1989,6 +1989,75 @@ export function useChatMessages(
     currentReasoningIdRef.current = null;
     currentToolCallIdRef.current = null;
 
+    // Strip interrupt segments populated by history replay before the reconnect stream
+    // re-delivers them; otherwise the same question/proposal renders twice (once on the
+    // history bubble, once on the reconnect bubble). The reconnect stream is
+    // authoritative for live interrupt state. Also redirect any unresolved-interrupt
+    // refs to point at the new reconnect bubble so the tool_call_result history-resolver
+    // writes resolution status to where the proposal actually lives now.
+    const stripList = unresolvedHistoryInterruptRef.current;
+    if (stripList && stripList.length > 0) {
+      const stripsByMsgId = new Map<string, HistoryInterruptInfo[]>();
+      for (const info of stripList) {
+        const arr = stripsByMsgId.get(info.assistantMessageId) || [];
+        arr.push(info);
+        stripsByMsgId.set(info.assistantMessageId, arr);
+      }
+      setMessages((prev) =>
+        prev.map((m) => {
+          if (m.role !== 'assistant') return m;
+          const strips = stripsByMsgId.get(m.id);
+          if (!strips) return m;
+          const msg = m as AssistantMessage;
+          const stripQuestionIds = new Set(
+            strips.filter((s) => s.type === 'ask_user_question' && s.questionId).map((s) => s.questionId!),
+          );
+          const stripProposalIds = new Set(strips.filter((s) => s.proposalId).map((s) => s.proposalId!));
+          const stripPlanApprovalIds = new Set(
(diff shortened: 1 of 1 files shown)


Chen-zexi:
On #2 (dead truthiness guard): applied in the latest commit.

On #1 (strip re-runs on retry): traced the retry path and the re-run is actually the intended behavior, not a bug.

`attemptReconnectAfterDisconnect` retries because the reconnect stream disconnected mid-delivery. Each retry opens a fresh `A_reconnect_Tn` bubble, and the Redis cursor still points past the original history watermark — so the buffered `interrupt` event is re-delivered into `A_reconnect_Tn` again. Without the re-strip, the previous reconnect bubble (`Tn-1`) AND the new one (`Tn`) would both end up holding the interrupt card → same duplicate, just shifted from history-vs-reconnect to reconnect-vs-reconnect.

What the second pass does:
- `stripsByMsgId` keys on the *current* `assistantMessageId` of each ref entry, which after the first redirect is `Tn-1` — exactly where the interrupt card was added by the first attempt's stream processor.
- Stripping `Tn-1` removes the card the first attempt populated; redirecting refs to `Tn` is what lets the `tool_call_result` resolver (line ~3236) write resolution status to the bubble where the proposal now actually lives after re-delivery.

The bot's suggested "clear stripList after first pass" would break that — the resolver iterates `unresolvedHistoryInterruptRef.current` to match resolutions, so an empty list means resolution status updates silently no-op. The proposal card would render as pending indefinitely until the user took an action.

End-state across retries:
- History bubble `H`: stripped of interrupt content on the first call ✓
- Earlier reconnect bubble `Tn-1`: stripped of interrupt content on the retry ✓
- Current reconnect bubble `Tn`: holds the (single, live) interrupt card ✓
- Resolver target: `Tn` ✓

The only concrete worry is a brief render window where `Tn-1` has been stripped but `Tn` hasn't yet received the re-delivery; that's a sub-frame flash and not user-visible. Happy to add an `attemptStrippedFor: Set<string>` guard if we hit a real symptom from this, but the current behavior matches what the dedup is supposed to do.

PR #218, opened 2026-05-20
feat(user-data): replace sandbox sync pipeline with virtual JSON backend

Chen-zexi:
## Summary

**Backend feature.** Mount `UserDataBackend` at `.agents/user/profile/{portfolio,watchlist,preference}.json`. Reads dispatch to `user_data_io` against live Postgres; writes diff against the current row set and apply atomically under `pg_advisory_xact_lock` + content-hash CAS. The `_PROTECTED_USER_FILES` guard in `file_ops` is gone — writes are validated by the backend itself. PTC drops the structured `user-profile` skill tools (flash keeps them, no filesystem there). `Write`/`Edit` on `.json` files now accept dict args directly via the tool-argument-parsing shim.

**Backend refactor.** Delete `sync_user_data.py` and every `mark_user_data_stale` / `_schedule_sync` call site across `workspace_manager`, `session_manager`, the users/portfolio/watchlist routers, and the user-profile skill tools. The agent now reads fresh JSON on demand, so the per-turn `.md` upload + invalidation plumbing (plus the `sandbox_user_data_upload_duration_ms` metric and the related hotpath check) is dead code — 1,095 lines gone for 58 added.

**Backend tests.** Unit coverage for `user_data_io` (serialize, parse, diff, version-conflict), `UserDataBackend` (read dispatch, cache, write validation), and the dict→JSON argument shim. Integration suite runs against real Postgres: insert/update/delete round-trips, advisory-lock serializing parallel writers, `other_preference` preservation on update, and end-to-end `aread_text` via the backend. Existing test files dropped now-removed sync-pipeline assertions.

**Frontend polish.** Route `profileRead` / `profileWrite` tool events through `agentPaths` + `toolDisplayConfig` so the activity strip shows entity-aware labels ("Read portfolio", "Updated preferences") instead of raw filenames. Pluralize `Preference → Preferences` in `en-US` and `zh-CN`. Memoize `currentCompletedIds`, `newlyCompletedIds`, and `summaryFingerprint` in `ActivityBlock` so the streaming list stops recomputing on every parent render. Add `aria-label` to the "open in tab" buttons in `ToolCallRow` + `EditToolRow`.

## Diff Size

- Total: 41 files changed, 4,256 insertions(+), 1,467 deletions(-)
- Net (excl. tests): 28 files changed, 2,414 insertions(+), 1,378 deletions(-)

## Test Coverage

`tests/integration/test_user_data_backend.py` (new, 8 cases against real Postgres):
- Portfolio insert/update/delete round-trips
- Version-conflict raised when payload version is stale
- Advisory lock serializes parallel writers — exactly one wins, one gets `version_conflict`
- Preferences first-insert seeds `other_preference = {}`
- Update preserves server-managed `other_preference` (the agent can't touch it)
- Watchlist rename = delete + insert by `name` (new DB id)
- End-to-end `UserDataBackend.aread_text` via the agent-facing surface

`tests/unit/server/services/test_user_data_io.py` (new) + `tests/unit/ptc_agent/agent/backends/test_user_data_backend.py` (new) + `tests/unit/ptc_agent/agent/middleware/tool/test_argument_parsing.py` (new) cover the pure functions, backend dispatch, cache, and the dict→JSON shim.

## Pre-Landing Review

Multi-specialist review (6 specialists) completed with `status="clean"` and quality score 9.5. One critical finding (`normalize_symbol`/`normalize_instrument_type` not applied on the agent-write path, causing two-writers-one-contract drift) and 50 informational findings. Critical resolved; 17 informational applied. Key fixes:

- `normalize_symbol` / `normalize_instrument_type` now applied in `user_data_io.py` so agent and dashboard agree on row identity
- `first_purchased_at` time component stripped in `serialize_portfolio` + matching `_portfolio_row_differs` change — no phantom diffs from `TIMESTAMPTZ` canonicalization
- `_locked_version_check` async context manager extracted to DRY the advisory-lock + content-hash recheck pattern across all three appliers
- Information-disclosure fix: stripped raw `{exc}` from the HTTP 500 response in `workspace_files.py`

## Design Review

Frontend changes are limited to label/locale/perf/a11y polish — no new components, no layout shifts. Lite design check ran during /review with clean output.

## Verification

- 186 unit tests pass on `HEAD`
- 8 new integration tests pass against real Postgres (Docker-local)
- 456 frontend tests pass on the merged state
- `ruff check` clean on all touched files (1 pre-existing F821 on `workspace_manager.py:184` for `PTCSandbox` forward-ref exists on `main` — unrelated to this PR)

## Test plan

- [x] Unit + integration tests green on `HEAD`
- [x] Lint clean on touched files
- [ ] Manual smoke: dashboard add a holding → agent reads `portfolio.json` → sees the new row
- [ ] Manual smoke: agent edits a holding's quantity via `Edit` → dashboard widget reflects the change after refresh
- [ ] Manual smoke: rename a watchlist via the agent — old name disappears, new name shows with items intact


Chen-zexi:
Both fixed in 90868180.

**1. `__version__` leak** — `_serialize_user_profile_file` now mirrors `UserDataBackend._read_serialized`:

```python
visible = {k: v for k, v in payload.items() if k != "__version__"}
return user_data_io.serialize_json(visible)
```

**2. Honest `truncated` flag** — computed from the actual slice:

```python
if unlimited:
    content = text_content
    truncated = False
else:
    lines = text_content.splitlines()
    content = "\n".join(lines[offset : offset + limit])
    truncated = len(lines) > offset + limit
```

Kept the 2-line strip inline rather than extracting a `_visible_payload` helper — two callers with the same pattern isn't enough surface to abstract. If a third caller shows up, that's the time.

PR #230, opened 2026-05-27
fix(llms): guard codex-oauth null Responses output

Chen-zexi:
## Summary

Fixes `TypeError: 'NoneType' object is not iterable` that kills codex-oauth turns.

The chatgpt.com Codex backend (`codex-oauth` provider, `use_response_api=true`) ships `response.output = null` on terminal `response.completed` stream frames. `langchain_openai` (1.2.2, latest) iterates that field unguarded in `_construct_lc_result_from_responses_api` (`chat_models/base.py`), so the whole turn dies.

- **fix(llms):** import-time guard in `ChatCodexOpenAI` that coerces `response.output` to `[]` before iteration, at the single chokepoint all read paths (sync/async/streaming) funnel through. Streamed text is already captured from `output_text.delta` events, so an empty terminal output loses no content. The guard is wrapped in `try/except (ImportError, AttributeError)` so a future langchain rename degrades to no-guard + a warning instead of breaking module import, and logs at debug when it fires.
- **test(llms):** regression test that runs a `output=None` Response through the real crash path (fails without the guard) + idempotency check.

Why not the request-side `exclude=["output"]` approach (openai codex-rs `d8044aede8`): the live backend rejects it with `HTTP 400 {'detail': 'Unsupported parameter: exclude'}` — it's only on OpenAI's dev branch, not prod.

## Diff Size
- Total: 2 files changed, 83 insertions, 1 deletion
- Net (excl. tests): 1 file changed, 45 insertions

## Test Coverage
`TestNullOutputGuard`: 2 tests — null output no longer crashes `_construct_lc_result_from_responses_api`; guard installs idempotently. Existing `ChatCodexOpenAI` tests preserved.

## Pre-Landing Review
2 robustness findings, both auto-fixed: (1) import-time hard dependency on a langchain internal could brick codex-oauth → wrapped in try/except; (2) silent coercion → added debug log. No SQL/shell/race/trust-boundary issues.

## Test plan
- [x] `tests/unit/llms/test_codex_extension.py` — 8/8 pass
- [x] Broader llms + codex-touching suite — 434 pass
- [x] ruff clean
- [x] Confirm against a live codex-oauth chat that previously crashed (client-side fix, cannot 400)

### DIFF
--- a/src/llms/extension/codex.py
+++ b/src/llms/extension/codex.py
@@ -128,6 +128,9 @@ def _install_responses_output_guard() -> None:
     patch the single chokepoint all read paths (sync, async, streaming) funnel
     through. Streamed text is already captured from ``output_text.delta`` events,
     so an empty terminal output loses no content.
+
+    Remove once langchain_openai guards ``response.output`` upstream — the guard
+    then becomes a no-op (it only mutates when ``output`` is None).
     """
     try:
         import langchain_openai.chat_models.base as _base
@@ -152,6 +155,9 @@ def _install_responses_output_guard() -> None:
             try:
                 response.output = []
             except Exception:
+                # Response isn't frozen today, so the line above succeeds. Kept as
+                # insurance: a frozen SDK model would raise pydantic ValidationError
+                # (not TypeError), and object.__setattr__ bypasses the frozen check.
                 object.__setattr__(response, "output", [])
         return orig(response, *args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


Chen-zexi:
**Note 1 — the narrowing would break the fallback, keeping `except Exception`.**

`Response` isn't frozen today (`model_config frozen=False`), so the primary `response.output = []` always succeeds and the fallback is dead code — agreed on that. But its purpose is a hypothetical *frozen* future `Response`, where assignment raises `pydantic_core.ValidationError` — not `TypeError`/`AttributeError`. Narrowing to `(TypeError, AttributeError)` would miss it and crash, defeating the fallback. `object.__setattr__` is the correct frozen-bypass. Verified against pydantic 2.13.4. Kept the broad except and added a comment explaining why.

**Note 2 — done.** Added a docstring line to remove the guard once `langchain_openai` guards `response.output` upstream.

Both in 392a860e.

PR #232, opened 2026-05-29
refactor(sse): drop asyncio.shield in workflow runner

Chen-zexi:
## Summary

Drops `asyncio.shield(inner_task)` from the background workflow runner, now that the Redis Streams cutover means no SSE consumer holds a reference to the workflow task.

- **Backend** (`background_task_manager.py`): replace `await asyncio.shield(inner_task)` with `await inner_task`; rename `_run_workflow_shielded` → `_run_workflow` (3 call sites); refresh the docstring to state the lifecycle is driven solely by `cancel_event` / `soft_interrupt_event`.
- **Tests**: add `test_outer_task_cancel_propagates_to_inner` proving `outer_task.cancel()` now propagates through `await inner_task`, cancels the inner consume loop, runs `_mark_cancelled`, and closes the workflow generator's `finally` block.

## Diff Size
- Total: 2 files changed, 97 insertions(+), 12 deletions(-)
- Net (excl. tests): 1 file changed, 10 insertions(+), 5 deletions(-)

## Behavioral change worth noting
With the shield gone, the two `info.task.cancel()` call sites — `shutdown()` force-cancel (line ~364) and `_cleanup_abandoned_tasks()` (line ~423) — now actually unwind the inner workflow, run `_mark_cancelled`, and close the generator. Pre-PR, the shield silently swallowed both, leaving orphaned generators (the cleanup function's name was a lie). This is a correctness fix. Dispatched runs left untouched for `abandoned_workflow_timeout` (6h default, `config.yaml`) now terminate cleanly instead of leaking.

## Test Coverage
The diff adds a regression test for the exact behavior change. `active_connections` tracking (which gates `_cleanup_abandoned_tasks`) is still maintained by the SSE handler at `stream_from_log.py:309/312`, verified during review.

## Pre-Landing Review
Clean. Ran `/review` on this diff — the only finding was a stale-base `.gitattributes` regression of #225, resolved by rebasing onto current `origin/main`.

## Test plan
- [x] Full unit suite: 3481 passed, 0 failures
- [x] `ruff check` on changed file: clean
- [x] e2e chat lifecycle sweep T1–T14: 14/14 pass (prior session)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation propagation for background workflows.
  * Enhanced tracking to accurately distinguish user-initiated cancellations from system-initiated cancellations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ginlix-ai/LangAlpha/pull/232?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/server/services/background_task_manager.py
+++ b/src/server/services/background_task_manager.py
@@ -1777,6 +1777,10 @@ class BackgroundTaskManager:
             task_info.status = TaskStatus.CANCELLED
             task_info.completed_at = datetime.now()
             metadata = task_info.metadata
+            # Only user-driven cancels set explicit_cancel; system force-cancels
+            # (shutdown timeout, abandoned-task cleanup) reach here via task.cancel()
+            # with the flag unset and must not be persisted as user actions.
+            cancelled_by_user = bool(task_info.explicit_cancel)
 
         logger.debug(f"[BackgroundTaskManager] Marked as cancelled: {key}")
 
@@ -1814,7 +1818,7 @@ class BackgroundTaskManager:
                     "agent_llm_preset": metadata.get("agent_llm_preset", "default"),
                     "deepthinking": metadata.get("deepthinking", False),
                     "is_byok": metadata.get("is_byok", False),
-                    "cancelled_by_user": True,
+                    "cancelled_by_user": cancelled_by_user,
                 }
 
                 await persistence_service.persist_cancelled(
--- a/src/server/services/background_task_manager.py
+++ b/src/server/services/background_task_manager.py
@@ -515,7 +515,7 @@ class BackgroundTaskManager:
                     existing.completion_callback = completion_callback
                     existing.graph = graph
                     existing.task = asyncio.create_task(
-                        self._run_workflow_shielded(
+                        self._run_workflow(
                             thread_id, run_id, workflow_generator,
                             cancel_event=existing.cancel_event,
                             soft_interrupt_event=existing.soft_interrupt_event,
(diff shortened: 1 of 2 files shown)


Chen-zexi:
Both findings verified against the code. Both fixed.

**1. `cancelled_by_user` hardcoded `True` (Major) — valid, fixed in dc3f0191.**
`_mark_cancelled` now reads the existing `task_info.explicit_cancel` flag instead of hardcoding `True`. Only the user-driven paths (`cancel_workflow`, `cancel_stale_workflow`) set that flag; the abandoned-task cleanup path (`task.cancel()` with no flag) was the one true mislabel and now persists `cancelled_by_user=False`.

Two clarifications on scope:
- This is **pre-existing**, not introduced by the shield removal. Per `asyncio.shield` semantics the awaiter still receives `CancelledError`, so the `except` at line 622 fired on outer-task cancel before this PR too. The shield only protected the *inner* task from cancellation, not the outer coroutine from the exception.
- `cancelled_by_user` is written once into `conversation_responses.metadata` and read nowhere else in `src/` (not billing, not control flow), so the blast radius was mislabeled analytics, not behavior. Fixed regardless.

**2. Closure-events test can false-pass (Minor) — valid, fixed in ac371eaa.**
Tightened to `pytest.raises(asyncio.CancelledError)` + `_mark_cancelled.assert_awaited_once_with(...)` + `_mark_completed.assert_not_awaited()`. It can no longer pass if `_run_workflow` ignored the cancel_event and ran to completion.

Added `TestMarkCancelledUserLabeling` (force-cancel → `False`, user cancel → `True`) as a regression guard for finding 1. Full BTM suite: 12 passed.

PR #245, opened 2026-06-02
feat(chat): add follow-up suggestion buttons after assistant replies

DOUBLES-J:
## 功能

聊天回复下方新增追问建议按钮：
- Settings → 模型 tab 新增开关控制（默认开启）
- 每次 AI 回复后自动生成 1-5 条追问建议（使用 resolve_model_client 解析凭证，支持 BYOK/OAuth）
- 点击建议按钮直接发送下一轮消息
- 通过 Redis Stream 推送 SSE 事件，零额外延迟

## 文件变更

| 文件 | 变更 | 说明 |
|------|------|------|
| `src/server/models/suggestion.py` | 新建 | Pydantic schema |
| `src/server/services/suggestion_service.py` | 新建 | 生成逻辑 + Redis 推送 |
| `src/server/app/threads.py` | 修改 | API 端点 |
| `src/server/handlers/chat/flash_workflow.py` | 修改 | completion callback |
| `src/server/handlers/chat/ptc_workflow.py` | 修改 | completion callback |
| `web/src/pages/ChatAgent/components/SuggestionButtons.tsx` | 新建 | UI 组件 |
| `web/src/pages/ChatAgent/components/MessageList.tsx` | 修改 | 集成渲染 |
| `web/src/pages/ChatAgent/components/ChatView.tsx` | 修改 | 点击回调 |
| `web/src/pages/ChatAgent/hooks/useChatMessages.ts` | 修改 | SSE handler |
| `web/src/pages/Settings/Settings.tsx` | 修改 | 开关 toggle |
| `web/src/types/chat.ts` | 修改 | 类型扩展 |
| `web/src/locales/*.json` | 修改 | i18n |
| `docs/api/01-plan/suggestion-buttons/PLAN.md` | 新建 | 实现计划 |

## 演示

演示视频见 PR 附件。

https://github.com/user-attachments/assets/266aa7c9-eaab-44c4-970d-a03c40de7e00

## Summary by CodeRabbit

* **New Features**
  * Added follow-up suggestion buttons that appear after AI replies, enabling quick follow-up interactions.
  * Added a Settings toggle to enable or disable follow-up suggestions.
  * Users can click suggestion buttons to instantly send them as chat messages.

* **Documentation**
  * Added comprehensive implementation plan for the suggestion buttons feature.

### DIFF
--- a/src/server/app/threads.py
+++ b/src/server/app/threads.py
@@ -1038,6 +1038,49 @@ async def get_thread_turns(thread_id: str, x_user_id: CurrentUserId):
     return await _get_thread_turns(thread_id, branch_tip_checkpoint_id=branch_tip)
 
 
+@router.get("/{thread_id}/turns/{run_id}/suggestions")
+async def get_turn_suggestions(
+    thread_id: str, run_id: str, x_user_id: CurrentUserId
+):
+    """Get follow-up suggestions for a completed turn.
+
+    Returns suggestions generated after the assistant reply was completed.
+    The suggestions are extracted from the ``sse_events`` JSONB column.
+    """
+    await require_thread_owner(thread_id, x_user_id)
+    from src.server.database.conversation import get_db_connection
+    from psycopg.rows import dict_row
+    import json as _json
+
+    async with get_db_connection() as conn:
+        async with conn.cursor(row_factory=dict_row) as cur:
+            await cur.execute(
+                """
+                SELECT sse_events
+                FROM conversation_responses
+                WHERE conversation_response_id = %s
+                  AND conversation_thread_id = %s
+                """,
+                (run_id, thread_id),
+            )
+            row = await cur.fetchone()
+
+    suggestions: list[str] = []
+    if row and row.get("sse_events"):
+        sse_events = row["sse_events"]
+        # sse_events may be a list or a JSON-string; normalize.
+        if isinstance(sse_events, str):
+            sse_events = _json.loads(sse_events)
+        if isinstance(sse_events, list):
+            for evt in sse_events:
+                if isinstance(evt, dict) and evt.get("event") == "suggestions":
+                    data = evt.get("data") or {}
+                    suggestions = data.get("suggestions") or []
+                    break
+
+    return {"suggestions": suggestions}
+
+
 @router.post("/{thread_id}/retry")
 async def retry_thread(
     thread_id: str,
--- a/src/server/handlers/chat/flash_workflow.py
+++ b/src/server/handlers/chat/flash_workflow.py
@@ -450,6 +450,46 @@ async def astream_flash_workflow(
                         sse_events=_sse_events,
                     )
 
+                # Generate follow-up suggestions BEFORE mark_completed so they
+                # are persisted before the SSE stream ends and the frontend
+                # fetches them via the suggestions API.
+                if _sse_events and config and user_id:
+                    try:
+                        from src.server.services.suggestion_service import (
+                            build_suggestion_sse_event,
+                            extract_assistant_text,
+                            generate_suggestions,
+                            push_suggestions_to_redis,
+                        )
+                        from src.server.database.conversation import update_sse_events
+
+                        assistant_text = extract_assistant_text(_sse_events)
+                        suggestions = await generate_suggestions(
+                            user_id=user_id,
+                            agent_config=config,
+                            assistant_text=assistant_text,
+                            call_logger=logger,
+                        )
+                        if suggestions:
+                            sse_event = build_suggestion_sse_event(suggestions)
+                            _sse_events.append(sse_event)
+                            await update_sse_events(run_id, _sse_events)
+                            # Push to Redis so the frontend receives the event
+                            # in the live SSE stream (no extra HTTP round-trip).
+                            await push_suggestions_to_redis(
+                                thread_id, run_id, suggestions,
+                            )
+                            logger.info(
+                                "[FLASH_COMPLETE] Suggestions appended: "
+                                f"thread_id={thread_id} count={len(suggestions)}"
+                            )
+                    except Exception:
+                        logger.warning(
+                            "[FLASH_COMPLETE] Suggestion generation failed, "
+                            "continuing",
+                            exc_info=True,
+                        )
+
                 await tracker.mark_completed(
                     thread_id=thread_id,
                     metadata={
--- a/src/server/handlers/chat/ptc_workflow.py
+++ b/src/server/handlers/chat/ptc_workflow.py
@@ -878,6 +878,46 @@ async def astream_ptc_workflow(
                     sse_events=_sse_events,
                 )
 
+                # Generate follow-up suggestions BEFORE mark_completed so they
+                # are persisted before the SSE stream ends and the frontend
+                # fetches them via the suggestions API.
+                if _sse_events and config and user_id:
+                    try:
+                        from src.server.services.suggestion_service import (
+                            build_suggestion_sse_event,
+                            extract_assistant_text,
+                            generate_suggestions,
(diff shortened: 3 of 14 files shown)


src/server/handlers/chat/flash_workflow.py
```diff
@@ -450,6 +450,46 @@ async def on_flash_workflow_complete(task_info):
...
+                        from src.server.services.suggestion_service import (
+                            build_suggestion_sse_event,
+                            extract_assistant_text,
+                            generate_suggestions,
+                            push_suggestions_to_redis,
+                        )
+                        from src.server.database.conversation import update_sse_events
+
+                        assistant_text = extract_assistant_text(_sse_events)
+                        suggestions = await generate_suggestions(
```
Chen-zexi, line 467: Move the `await`ed `generate_suggestions` to **after** `mark_completed()` and run it detached (fire-and-forget). Hold a reference to the task so it isn't garbage-collected, and capture the locals it needs (`handler` / `sse_events`) before the manager's terminal-ref cleanup nulls them. Right now the turn can't close until this flash call returns. (Same placement in `ptc_workflow.py` — note there too.)

src/server/handlers/chat/ptc_workflow.py
```diff
@@ -878,6 +878,46 @@ async def on_background_workflow_complete(task_info):
...
+                        from src.server.services.suggestion_service import (
+                            build_suggestion_sse_event,
+                            extract_assistant_text,
+                            generate_suggestions,
+                            push_suggestions_to_redis,
+                        )
+                        from src.server.database.conversation import update_sse_events
+
+                        assistant_text = extract_assistant_text(_sse_events)
+                        suggestions = await generate_suggestions(
```
Chen-zexi, line 895: Same as the flash callback: this `await`ed `generate_suggestions` sits before `mark_completed()` on the completion-critical path. Move it after `mark_completed()` as a detached fire-and-forget task.

src/server/services/suggestion_service.py
```diff
@@ -0,0 +1,241 @@
...
+        log.warning("[SUGGESTION] No flash model available, skipping")
+        return []
+
+    # --- Resolve credentials ---------------------------------------------------
+    # Use resolve_model_client (OAuth → BYOK → platform fallback) instead of
+    # create_llm so BYOK / custom provider keys stored in user preferences
+    # are discovered. Avoids resolve_llm_config → _cow → model_copy(deep=True)
+    # which can fail on mutated global config objects.
+    try:
+        resolved = await resolve_model_client(
```
Chen-zexi, line 100: Use the canonical wrapper instead — `resolve_model_client` is the internal primitive of `resolve_llm_config`, and going around `LLMService.complete` skips fallback/subsidiary resolution and the `platform_key_fallback` billing log:

```python
from src.server.app import setup
result = await setup.llm_service.complete(
    user_id=user_id, mode="flash",
    system_prompt=SUGGESTION_SYSTEM_PROMPT,
    user_prompt=_build_user_prompt(assistant_text),
    response_schema=SuggestionResponse,
)
```

It must run on the **pristine** `setup.agent_config` (which `LLMService` already holds), not the workflow's resolved `config`. `complete()` resolves the user's preferred flash model per-call, so BYOK/OAuth still applies.

src/server/services/suggestion_service.py
```diff
@@ -0,0 +1,241 @@
...
+Generates 1-5 follow-up questions after each assistant reply, respecting
+the user's ``suggestion_enabled`` preference (default: enabled). Results
+are appended to the ``sse_events`` JSONB column as a ``suggestions`` event
+so the frontend can retrieve them via the dedicated suggestions API endpoint.
+
+Model selection: reads the user's ``preferred_flash_model`` from
+``other_preference``, falls back to ``agent_config.llm.flash``.
+Credential resolution uses ``resolve_model_client`` (OAuth → BYOK →
+platform fallback) so BYOK / custom provider keys are discovered. This
+avoids ``resolve_llm_config`` → ``_cow`` → ``model_copy(deep=True)``
```
Chen-zexi, line 12: This attribution is wrong — worth fixing so it doesn't mislead. The `RLock` isn't injected by middleware/LangGraph; it rides in on the `llm_client` that `resolve_llm_config` itself builds (anthropic SDK → httpx). Deep-copying the **pristine** config is always fine (the unconditional `_cow()` runs clean) — it only crashes when you deep-copy an already-resolved config whose client has been **used**. So `LLMService.complete` on the pristine config works; the bypass was solving a self-inflicted problem.

src/server/services/suggestion_service.py
```diff
@@ -0,0 +1,241 @@
...
+    except Exception:
+        logger.warning("[SUGGESTION] Redis unavailable, cannot push to stream")
+        return False
+
+    if not cache.enabled:
+        return False
+
+    data = {"suggestions": suggestions}
+    sse_str = (
+        f"id: 0\nevent: suggestions\n"
```
Chen-zexi, line 191: Drop the live push. On the **PTC path**, by the time this runs `clear_event_buffer` has already `DEL`'d `workflow:stream:{thread_id}:{run_id}` (`ptc_workflow.py` wires `_on_pair_persisted = clear_event_buffer`, run inside `persist_completion`), so the `XADD` re-creates the stream with just this one event — hence the hardcoded `id: 0`. On both paths it also races the consumer, which can already have exited on the in-memory `COMPLETED` status. Let the client fetch after stream-close instead.

src/server/app/threads.py
```diff
@@ -1038,6 +1038,49 @@ async def get_thread_turns(thread_id: str, x_user_id: CurrentUserId):
...
+    await require_thread_owner(thread_id, x_user_id)
+    from src.server.database.conversation import get_db_connection
+    from psycopg.rows import dict_row
+    import json as _json
+
+    async with get_db_connection() as conn:
+        async with conn.cursor(row_factory=dict_row) as cur:
+            await cur.execute(
+                """
+                SELECT sse_events
```
Chen-zexi, line 1059: Back this with a durable **`suggestions` JSONB column on `conversation_responses`** plus a Redis key for in-flight state (`200` ready / `202` pending / column fallback on expiry). Today suggestions live only in `sse_events`, and the client's history-replay path doesn't apply the `suggestions` event, so they vanish on refresh — a durable column behind this endpoint makes it the single reliable source.

Chen-zexi (commented):
Thanks for the PR. The core concern: follow-up suggestions are a best-effort, non-critical feature, but the implementation runs the suggestion LLM call synchronously on the turn's completion-critical path — in the completion callback, before `mark_completed()`. Two reasons we're asking for changes:

1. **A non-critical service is gating the critical path.** The suggestion LLM call is `await`ed before completion proceeds, with no timeout around it — so a slow or hung call holds up `mark_completed`, the manager's post-callback burst-slot release and ref cleanup, and stream teardown for as long as the model takes (the SDK read timeout is 600s, and structured-output parsing retries up to 5×). A best-effort, cosmetic enhancement shouldn't be able to hold up the core turn lifecycle. (The same placement also makes the live push race stream teardown — on the PTC path it `XADD`s into a stream `clear_event_buffer` has already deleted — and the chips don't survive a refresh; both downstream of this.)
2. **It bypasses the shared LLM primitive.** It reaches past `LLMService.complete` into the internal `resolve_model_client`, skipping fallback/subsidiary resolution and the `platform_key_fallback` billing log — and the comment justifying the bypass misdiagnoses why: the `RLock` comes from the resolver's own client, not middleware, and `LLMService.complete` on the pristine config doesn't crash.

Requests: (1) move generation off the critical path (fire-and-forget after `mark_completed`), (2) use `setup.llm_service.complete` on the pristine config, (3) drop the live Redis push for a fetch-after-close, (4) store durably. Details inline.

PR #252, opened 2026-06-09
Separate portfolio NAV by currency

### LINKED ISSUE #189: In the current project, the calculation of the total portfolio value does not distinguish between different currency types and converts everything to USD.

gyx09212214-prog:
## Summary
- carry each holding's currency through portfolio rows instead of dropping it in the dashboard hook
- group portfolio NAV, cost basis, and unrealized P/L by currency in the main portfolio card and dashboard widgets
- render portfolio row prices/market values with the holding currency instead of hard-coded USD
- include currency-aware NAV lines and row currency in widget context exports
- add regression tests for multi-currency summary behavior

Fixes #189.

## Tests
- `corepack pnpm exec vitest run src/pages/Dashboard/utils/__tests__/portfolioSummary.test.ts`
- `corepack pnpm exec vitest run src/locales/__tests__/keys.test.ts src/pages/Dashboard/utils/__tests__/portfolioSummary.test.ts`
- `corepack pnpm exec tsc --noEmit`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Multi-currency portfolio: per-currency NAV totals, per-currency P/L badges, locale-aware currency formatting, and snapshot/markdown export that includes currency columns when relevant; hidden-value masking preserved.
  * Improved handling of missing price data so unavailable quotes don’t show misleading values; extended-hours info only shown when present.

* **Localization**
  * Added English and Chinese labels for “Net Asset Value by Currency”.

* **Tests**
  * Added unit tests for currency normalization, per-currency summaries, formatting, and missing-quote handling.


gyx09212214-prog:
Automated review follow-up pushed in `6d01e37c`.

Handled/verified CodeRabbit feedback:
- Confirmed the Chinese `navByCurrency` / `netAssetValueByCurrency` wording is aligned as `按币种净资产价值`.
- Confirmed portfolio summary calculation is memoized with `React.useMemo`.
- Confirmed `PortfolioRow.currency` is typed as `string` to match runtime normalization.
- Confirmed zero-value NAV summaries are filtered before display.
- Confirmed percentage formatting uses `createFormatter` in snapshot markdown.
- Tightened the remaining default-currency test rows to use the existing typed `createPortfolioRow` fixture.

Validation:
- `git diff --check`: passed
- `corepack pnpm exec vitest run src/pages/Dashboard/utils/__tests__/portfolioSummary.test.ts`: 3 passed
- `corepack pnpm test -- src/pages/Dashboard/utils/__tests__/portfolioSummary.test.ts`: passed in this environment

web/src/pages/Dashboard/utils/portfolioSummary.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export function summarizePortfolioByCurrency(rows: PortfolioSummaryRow[]): CurrencyPortfolioSummary[] {
+  const groups = new Map<string, { currency: string; totalValue: number; totalCost: number }>();
+
+  rows.forEach((row) => {
+    const currency = normalizePortfolioCurrency(row.currency);
+    const current = groups.get(currency) ?? { currency, totalValue: 0, totalCost: 0 };
+    const quantity = finiteNumber(row.quantity);
+    current.totalValue += finiteNumber(row.marketValue);
```
Chen-zexi: This is the line that collides with #254: `finiteNumber(null)` is `0`, so once #254 lands (it makes `marketValue` null when a quote is missing), an unpriced holding contributes 0 here while its cost still counts — on a test merge I got `totalValue: 0, totalCost: 50, totalPlPct: -100` for a holding whose quote was simply missing. After rebasing, filter before grouping:

```ts
const pricedRows = rows.filter(
  (r) => r.quoteAvailable !== false && r.marketValue != null
);
```

plus `quoteAvailable?: boolean` on `PortfolioSummaryRow`. Please also re-check the "NAV shows N/A when nothing is priced" path per currency group afterwards.

web/src/pages/Dashboard/components/PortfolioWatchlistCard.tsx
```diff
@@ -483,21 +487,35 @@ function PortfolioWatchlistCard({
...
                   </div>
                   <div
-                    className="text-2xl font-bold mb-2 dashboard-mono"
+                    className={`${portfolioSummaries.length > 1 ? 'text-xl' : 'text-2xl'} font-bold mb-2 dashboard-mono`}
                     style={{ color: 'var(--color-text-primary)' }}
                   >
-                    {valuesHidden ? '********' : `$${fmt2(totalValue)}`}
+                    {valuesHidden
+                      ? '********'
+                      : portfolioSummaries.map((summary) => (
```
Chen-zexi: This maps every currency group, but `PortfolioNavSummary` filters out `totalValue === 0` and shows `--`. A sold-out holding (qty 0) leaves a stray `0.00` NAV line here but not in the widget. Please apply the same filter in both spots.

web/src/pages/Dashboard/widgets/definitions/_holdingsPrimitives.tsx
```diff
@@ -192,12 +197,12 @@ export function PortfolioRowItem({
...
             {valuesHidden
               ? '******'
-              : `$${fmt2(Number(item.marketValue || 0))}`}
+              : formatPortfolioMoney(Number(item.marketValue || 0), currency, i18n.language)}
           </div>
           <div className="text-xs dashboard-mono" style={{ color: 'var(--color-text-secondary)' }}>
             {valuesHidden
               ? '***'
-              : `@${fmt2(Number(extType && item.previousClose != null ? item.previousClose : item.price))}`}
+              : `@${formatPortfolioMoney(Number(extType && item.previousClose != null ? item.previousClose : item.price), currency, i18n.language)}`}
```
Chen-zexi: This used to render `@182.31` and now renders `@$182.31` (or `@HK$800.00`) — the `@` next to the currency symbol reads doubled. Keep the bare localized number for this secondary line, or drop the `@`.

Chen-zexi (changes requested):
Thanks for the PR.

Critical, before this can merge: **this branch conflicts with #254 in five files, and as written `summarizePortfolioByCurrency` silently undoes #254's fix.** On a test merge of both branches, `finiteNumber(null)` turns a missing quote into 0, so an unpriced holding came out as `{currency: "CNY", totalValue: 0, totalCost: 50, totalPlPct: -100}` — the exact #248 bug, per-currency. None of the merge-time type errors point near it.

Plan: #254 lands first, then please:
1. Rebase this on top of it.
2. Filter unpriced rows before grouping (inline suggestion on the exact line).
3. Add a regression test: a multi-currency portfolio where one holding has `quoteAvailable: false`, asserting it appears in neither that currency's NAV nor its cost basis.
4. Fix the two inline nits (stray `0.00` NAV line, `@$` formatting).
5. Add tests for `formatPortfolioMoney` and the serializer's new `currency` column — both are currently only exercised through the `$`-fallback fixtures.

Chen-zexi:
Thanks for the updates — the inline asks here are all addressed.

The one thing left before merge is sequencing with #254. This PR now also implements the missing-quote handling itself (`quoteAvailable` + `marketValue: null` in `usePortfolioData.ts`, and `getStockPrices` dropping unpriced symbols in `api.ts`), which is the same ground #254 covers — so the two now conflict across the shared portfolio files.

We don't want to drop either: #254 is the broader of the two. It also fixes the watchlist path (`useWatchlistData`, `WatchlistWidget`, `AddWatchlistItemDialog`) and the `N/A` / `—` rendering for unpriced rows, which this PR doesn't touch.

Our preference is to land #254 first, since it owns the missing-quote contract end to end, then rebase this PR on top — at which point #252 can drop the now-redundant `quoteAvailable` / `getStockPrices` changes and keep just the currency grouping + the priced-row filter, which already consumes #254's row shape.

How would you like to handle it — does landing #254 first and rebasing this on top work for you, or did you have a different plan in mind?

gyx09212214-prog:
That works for me. Let's land #254 first since it owns the missing-quote contract across both portfolio and watchlist paths. Once #254 is merged, I'll rebase #252 on top, drop the redundant `quoteAvailable` / `marketValue: null` / `getStockPrices` handling from this branch, and keep the currency grouping plus the priced-row filter. I'll rerun the focused dashboard tests before pushing the rebased branch.

Chen-zexi:
Hey @gyx09212214-prog , #254 has landed, feel free to ping me once this PR is ready

gyx09212214-prog:
Thanks! I've rebased this PR on top of the latest main after #254 landed and resolved the conflicts. The branch is ready for another review.

Chen-zexi (approved):
Thanks for the update.

PR #253, opened 2026-06-09
Route batch snapshots by symbol market

gyx09212214-prog:
## Summary
- route batch stock snapshots through provider entries that cover each symbol market
- continue fallback when a provider returns no rows or only resolves part of a batch
- preserve requested symbol order for resolved snapshots
- add regression tests for mixed US/CN batches and empty-result fallback

Addresses #248.

## Tests
- `python -m pytest tests\unit\data_client\test_market_data_provider.py -q -k "not ConfigAccessor"`
- `git diff --check`

Attempted:
- `$env:PYTHONPATH='.;src'; python -m pytest tests\unit\data_client\test_market_data_provider.py -q`
  - local system Python is missing project dependency `structlog`, so the two existing `TestConfigAccessor` tests fail during config import; the 24 provider-routing tests pass.

Note: this fixes the backend batch snapshot routing path behind the A-share zero-price issue. The frontend missing-snapshot display fallback can be handled separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Market snapshot retrieval now routes requests per symbol across providers, preserves original symbol order, drops provider-returned rows missing or not matching requested symbols (with warnings), logs targeted fallbacks, returns empty when providers supported the call but none resolved, or surfaces the final error when providers fail.

* **Tests**
  * New unit tests for per-symbol routing, partial/full fallbacks, normalization, dropped/unkeyed rows, and error propagation.

### DIFF
--- a/src/data_client/market_data_provider.py
+++ b/src/data_client/market_data_provider.py
@@ -211,7 +211,7 @@ class MarketDataProvider:
             user_id=user_id,
         )
 
-    # -- Snapshot interface (batch; no per-symbol routing) --------------------
+    # -- Snapshot interface ---------------------------------------------------
 
     async def get_snapshots(
         self,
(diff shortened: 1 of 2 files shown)


gyx09212214-prog:
Automated review follow-up pushed in `24e2ee60`.

Handled comments:
- review by @coderabbitai: https://github.com/ginlix-ai/LangAlpha/pull/253#pullrequestreview-4456642313

Changes:
- Pushed previously validated local review-follow-up commit.

Validation:
- `previous Codex validation`: Passed before the earlier push attempt.

src/data_client/market_data_provider.py
```diff
@@ -211,30 +211,78 @@ async def get_daily_with_source(
...
+                continue
+
+            resolved: set[str] = set()
+            for snap in snapshots or []:
+                symbol = str(snap.get("symbol") or "").upper()
+                if symbol:
+                    results_by_symbol[symbol] = snap
+                    resolved.add(symbol)
+                else:
+                    extras.append(snap)
```
Chen-zexi: If a source returns a snapshot without a usable `symbol` key, it lands in `extras` but the symbol stays pending — the next provider is called for it too, so the same symbol can appear twice in the output (reproduced this with a scripted source: 1 requested symbol → 2 snapshots returned). The symbol-less extra then fails `SnapshotData(**item)` in `market_data.py:671` (`symbol` is required), turning the whole batch into a 500.

Please log a warning and drop these instead of collecting them — nothing downstream can use a snapshot it can't key, and that closes both seams in one line.

src/data_client/market_data_provider.py
```diff
@@ -211,30 +211,78 @@ async def get_daily_with_source(
...
+                for symbol in symbols
+                if str(symbol).upper() in results_by_symbol
+            ]
+            return ordered + extras
+
         if last_exc:
             raise last_exc
+        if tried_any:
+            return []
         raise RuntimeError("No data source supports get_snapshots")
```
Chen-zexi, line 299: If a symbol's market matches no entry and there's no `{all}` entry in the chain, `tried_any` stays false and this raises — even when other symbols in the batch resolved fine. Can't happen with today's config (fmp is `{all}`), but it'd bite a market-scoped-only setup. Please reserve the `RuntimeError` for "no entry implements `get_snapshots` at all" and otherwise return whatever resolved.

Chen-zexi (changes requested):
Thanks for the PR.

Two edge cases to address before merge — see the inline comments (the `extras` double-emit and the uncovered-market `RuntimeError`). Please also add tests for those paths: a source returning a snapshot without a `symbol` key, and all-providers-error re-raising `last_exc`.

Ignore the failing docstring-coverage check — it's counting your test helper.

gyx09212214-prog:
I re-checked the review feedback against the latest head `36c6993`.

The previously requested/blocking items appear addressed now:

- symbol-less snapshot rows are logged and dropped via `drop_unkeyed` instead of being returned downstream
- keyed rows that were not requested in the current provider batch are logged and dropped via `drop_unrequested`
- uncovered-market symbols no longer force a `RuntimeError` when other symbols resolve; the `RuntimeError` is now reserved for the case where no source implements `get_snapshots`
- whitespace-padded symbols use the same normalization for routing, pending removal, and final result lookup
- regression coverage is present for partial fallback, empty-result fallback, unkeyed/unrequested rows, uncovered-market partial results, all-provider errors, and whitespace-padded suffix symbols

I also verified locally:

```text
PYTHONPATH=.;src python -m pytest tests\unit\data_client\test_market_data_provider.py -q -k "not ConfigAccessor"
31 passed, 2 deselected

git diff --check
passed

Chen-zexi (approved):

Chen-zexi:
Thanks again for the PR! Merge it now

PR #255, opened 2026-06-09
feat(news): TickerTick dashboard news source + cache-stampede hardening

Chen-zexi:
## Summary

Adds **TickerTick** as a dashboard news source and hardens the whole `/news`
read path for production load. Four logical pieces:

**News feature (TickerTick)**
- New `TickerTickNewsSource` (keyless `api.tickertick.com`) wired into the data-client
  registry as a directly-addressable provider. Powers a new curated **"Top"** tab and
  re-points the Portfolio/Watchlist tabs at TickerTick's entity-aware ticker coverage.
- `/news` gains a validated `provider` query param; a by-source publisher filter and a
  detail modal round out the News widget. EN + zh-CN locales.

**Cache-stampede protection** (`3649c0ac`)
- Two-layer single-flight on cold-cache misses: in-process per `(provider,tickers,limit)`
  key, then a cross-worker `SET NX PX` Redis lock so exactly one worker fetches a cold key
  while the rest read the filled cache. Cursor/date/sort-filtered requests bypass the shared
  path. `get_article_by_id` now scans via non-blocking `SCAN` instead of `KEYS`.

**Keep-warm poller** (`35c211e8`)
- `NewsRefreshService`: a singleton background poller that delta-merges the latest page of
  the curated + Market general feeds into a rolling buffer (by id, newest-first, capped) under
  the same Redis key `/news` reads — global feeds stay warm and never cold-miss upstream. One
  worker polls per tick via a Redis leader lock. Config under `news_poll` in `config.yaml`.

**Hardening**
- TickerTick adapter (`0fd053d0`): non-dict upstream bodies degrade to an empty feed instead
  of crashing; ticker tokens are sanitized before entering the query DSL (injection guard).
- News modal (`a8a7380f`): article/fallback URLs routed through a `safeHttpUrl` guard that
  blocks `javascript:`/`data:` schemes before they reach an `<a href>`.

**Frontend cleanup** (`b6f8ee8b`)
- `useTickerNews` migrated off its hand-rolled module cache + `setInterval` onto React Query,
  sharing one cache/poll mechanism with `useDashboardData`. Duplicated mapper/formatter/types
  extracted to a shared `newsItem` module (fixes zh-CN relative-time on the portfolio/watchlist
  path). `getNews` re-throws on error so callers retry and retain the last good list instead of
  blanking the feed.

## Diff Size
- Total: 35 files changed, +2,545 / -232
- Net (excl. tests): 25 files changed, +1,191 / -232

## Test Coverage
- Backend: Redis lock/scan primitives, `/news` single-flight + distributed-lock paths,
  delta-merge poller, provider-scoped cache keys, TickerTick query-injection + timestamp guards.
- Frontend: `useTickerNews` React Query migration (eager fetch + 60s repoll, provider/cacheKey
  isolation, no-ticker short-circuit, unmount stops polling, field fallbacks, failed-poll
  retains last list).
- Full backend unit suite: **3670 passed**. Frontend Dashboard: **264 passed**.

## Pre-Landing Review
Focused delta review (this code was reviewed twice pre-merge). No functional bugs found.
One known gap:
- **`safeHttpUrl` XSS guard has no regression test.** The guard works, but `NewsDetailModal.test.tsx`
  has no scheme coverage — a future refactor could silently drop it. Trivial to cover; tracked as
  a follow-up.

## Verification
- [x] `uv run pytest tests/unit/` — 3670 passed
- [x] `pnpm vitest run src/pages/Dashboard` — 264 passed
- [x] `ruff check` (changed files) — clean
- [x] `tsc --noEmit` — clean
- [x] `eslint` (changed files) — clean

## Notes
- `agent_config.yaml` intentionally **not** included.
- Rebased onto `origin/main`; linear history, no conflicts.

## Summary by CodeRabbit

* **New Features**
  * "Top" curated news tab with infinite-scroll (warmed, de-duplicated results)
  * Optional provider selection for news requests (including a new curated provider)
  * Modal accepts row fallback so article details render instantly without a by-id fetch

* **Improvements**
  * Background poller keeps Top feed hot for faster loads and rolling buffer merges
  * Article list now inlines author, description, keywords and sentiment details
  * Safer article links via stricter URL validation; updated dashboard/localization strings

* **Tests**
  * Extensive unit tests covering polling, caching, locking, provider routing, and UI behavior

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -165,6 +165,22 @@ news_data:
     - name: fmp
     - name: yfinance
 
+# News refresh poller — keeps the GLOBAL feeds (curated "Top" + Market general)
+# warm by fetching the latest page every interval and delta-merging new articles
+# (by id) into a rolling buffer stored in the same Redis key the /news endpoint
+# reads. Per-ticker feeds are unbounded and stay on the on-demand cache. One
+# worker polls per tick (Redis-lock leader election); the buffer self-heals via
+# the cache TTL if the poller stops.
+news_poll:
+  enabled: true
+  interval_seconds: 60     # How often to fetch the latest page
+  max_items: 100           # Rolling-buffer cap per feed
+  feeds:
+    - provider: tickertick # Curated "Top" feed (news:tickertick:general:<limit>)
+      limit: 50
+    - provider: null       # Provider-chain general feed / Market tab (news:general:<limit>)
+      limit: 50
+
 
 # =============================================================================
 # MARKET INSIGHT CONFIGURATION
--- a/src/config/models.py
+++ b/src/config/models.py
@@ -180,7 +180,8 @@ class NewsPollFeedConfig(BaseModel):
 
     ``provider`` None targets the provider chain (Market general feed); a name
     (e.g. ``tickertick``) targets that source directly. Always polled with no
-    tickers, so it maps to a global ``news:[provider]:general:<limit>`` key.
+    tickers, so it maps to a global cache key — ``news:tickertick:general:50``
+    with a provider, ``news:general:50`` without one.
     """
 
     provider: str | None = Field(default=None)
--- a/src/config/models.py
+++ b/src/config/models.py
@@ -175,6 +175,27 @@ class NewsDataConfig(BaseModel):
     providers: List[MarketDataProviderConfig] = Field(default_factory=list)
 
 
+class NewsPollFeedConfig(BaseModel):
+    """One global feed kept warm by the news refresh poller.
+
+    ``provider`` None targets the provider chain (Market general feed); a name
+    (e.g. ``tickertick``) targets that source directly. Always polled with no
+    tickers, so it maps to a global ``news:[provider]:general:<limit>`` key.
+    """
+
+    provider: str | None = Field(default=None)
+    limit: int = Field(default=50, ge=1, le=100)
+
+
+class NewsPollConfig(BaseModel):
+    """News refresh poller — delta-merges the latest page into a rolling buffer."""
+
+    enabled: bool = Field(default=True)
+    interval_seconds: int = Field(default=60, ge=10)
+    max_items: int = Field(default=100, ge=1, le=500)
+    feeds: List[NewsPollFeedConfig] = Field(default_factory=list)
+
+
 class InfrastructureConfig(BaseModel):
     """Root model for infrastructure configuration (config.yaml)."""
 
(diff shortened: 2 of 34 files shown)


Chen-zexi:
**Re: CodeRabbit nitpicks** — applied the two actionable ones, deferring the locale item.

- **`models.py` `NewsPollFeedConfig` docstring (provider=None key form)** — applied in `5d32c5b3`. Now documents both `news:tickertick:general:50` (with provider) and `news:general:50` (without).
- **`settings.py` `get_news_poll_config()` return annotation** — applied in `5d32c5b3`. Now `-> NewsPollConfig`; verified no circular import (`models.py` has no back-import into `settings`/`core`).
- **`newsItem.ts` relative-time staleness on locale switch** — deferred. The string is cached in React Query and self-heals on the next 60s poll; moving `formatRelativeTime` to render time is a widget refactor for a ≤60s cosmetic lag. Tracked, not blocking.

Chen-zexi:
**Re: claude[bot] reviews** — verdict per finding across both passes.

**Applied**
- **Double JSON-encode in `NewsCacheService`** → `9f38541b`. `set` now hands the dict straight to `RedisCacheClient` (single encode); `get`/`get_article_by_id` drop the redundant `json.loads`. This pre-dates the branch (merge-base already had it). Deploy caveat: old double-encoded `news:*` entries read back as a string until their 3–5 min TTL expires — self-heals, or flush `news:*` after rollout for a zero-window cutover.
- **`acquire_lock`/`release_lock` missing the try/except guard** → `2f314c60`. Confirmed the failure mode: an unguarded `get_cache_client()` exception propagated through `_get_news_data` into the shared single-flight task, where `asyncio.shield` re-raised it to every concurrent waiter as a 500. They now return the safe default (None / no-op) so the leader falls back to a direct fetch. Regression test added.

**Declined**
- **`ensure_future` → `create_task`** — premise doesn't hold. `asyncio.ensure_future` is not deprecated on 3.12 (no `DeprecationWarning`, no docstring note), and `_single_flight`'s param is typed `Callable[[], Awaitable[dict]]`. `create_task` raises `TypeError` on any non-coroutine awaitable (verified: *"a coroutine was expected, got <Future...>"*), so `ensure_future` is the contract-correct choice here. Keeping as-is.

**Deferred**
- **`TickerTickClient` per-call construction (no cross-request connection reuse)** — low impact: the 60s warm-poller + 300/180s cache absorb nearly all upstream hits, and concurrent cold-miss bursts already collapse to one connection via single-flight. A shared client needs lifespan-managed shutdown and careful event-loop binding (a pytest footgun), so it's cleaner as its own change if/when ticker-feed miss rates pressure tail latency. Tracked.

(The `safeHttpUrl` scheme-rejection regression test remains a tracked follow-up, as noted.)

PR #263, opened 2026-06-12
feat(mcp): multi-tenant, per-workspace MCP server configuration

Chen-zexi:
## Summary

Lets users configure their own MCP (Model Context Protocol) servers per workspace — add, edit, enable/disable, delete, import, and promote-to-template — with secrets kept in the vault and untrusted server code confined to the user's own Daytona sandbox. Decouples MCP config from the static process-global `agent_config.yaml` registry into a DB-backed, per-workspace, versioned, runtime-resolved system.

**Backend — data + resolution**
- New tables in a single consolidated migration `012`: `user_mcp_servers` (catalog templates), `workspace_mcp_servers` (per-workspace source of truth), `workspace_mcp_tool_schemas` (discovery cache, keyed by a **per-server `config_hash`** so mutating one server doesn't orphan the rest), plus a `mcp_config_version` counter on `workspaces` bumped in-txn on every mutation.
- `resolve_mcp_config` chokepoint merges built-ins (minus disable-markers) + workspace servers; one shared resolver feeds both the prompt summary and sandbox sync, resolved once per session and cached.
- Composite registry: `SchemaOnlyRegistry` appended over the frozen built-in registry; host `connect_all` stays **built-ins-only** (SSRF boundary by construction).

**Backend — sandbox + secrets**
- In-sandbox tool discovery via a `discover()` entrypoint in the generated `mcp_client.py` (file-IPC, not stdout); discovery runs with **no vault access** unless `discovery_uses_secrets` is set.
- `${vault:NAME}` resolves **only** from the sandbox vault file, per-server-scoped, never host `os.environ`.
- Untrusted-schema sanitization: legal-identifier coercion + collision handling, tool/param text sanitized, workspace servers neutral-framed in the prompt (not under the authoritative `Instructions:` label), detailed-mode bounded by per-server caps.

**Backend — API**
- Catalog CRUD (`/api/v1/mcp/servers`) and per-workspace endpoints (effective list, add, edit, toggle, delete, discover probe, bulk JSON import, promote-to-template). Pydantic validation: name regex, transport↔field coherence, URL policy (https, no private/link-local/userinfo), command allowlist (no `bash`), length caps, vault-ref-only secret values.

**Frontend**
- New MCP tab (`This Workspace | Templates`): server rows with origin/status, add/edit modal with paste-to-fill + vault secret picker, bulk JSON import (auto-vaults credential-looking values), Saved→Verifying→Ready lifecycle indicator, save-as-template promotion.

## Diff Size
- Total: 65 files changed, +15,421 / −172
- Net (excl. tests): 39 files changed, +8,181 / −171

## Test plan
- [x] Backend unit suite — **4,098 passed**, 0 failures (`uv run pytest tests/unit/`)
- [x] MCP DB integration suite — **21 passed** against real Postgres (incl. schema-GC purge + no-downgrade tests)
- [x] Frontend — **1,346 passed** across 118 files, 0 failures (`pnpm vitest run`)
- [x] `ruff check src/` clean for the feature; ESLint 0 errors
- [x] Rebased onto latest `main` — 18 linear commits, no merge commit; the one conflict in `daytona.py` (uvx PATH vs. main's sandbox-image unification) re-resolved keeping both; branch patch verified byte-identical pre/post rebase, all suites re-run green
- [x] E2E against a live sandbox (add remote server + secret → discover → agent calls it; stdio cold-start; disable built-in → pruned) — recommended before/after merge via `/e2e-chat`

## Plan completion
24/24 planned deliverables DONE (all 6 iron-rule regression tests present). Two intentional improvements over the approved plan: discovery cache re-keyed from workspace-wide `config_version` → per-server `config_hash`; lifecycle/import/promote UI added beyond original scope.

## Known issues / review findings

Pre-landing review ran cross-model (Claude security + Claude adversarial + Codex) plus a dedicated performance/hot-path pass.

**Hot-path verification** — all five latency invariants verified in code: warm acquire inside the 30s cooldown does zero DB reads; the `mcp_config_version` check piggybacks the existing post-cooldown workspaces read (no new per-turn query); `create_agent` reads the session-cached composite registry + precomputed tool summary every turn; discovery is never awaited on the response path; zero-user-server sync stays a single-manifest-round-trip no-op. The only new chat-path cost is the first turn after a mutation (~2 PG round trips + bounded asset sync), which the proactive apply absorbs at mutation time.

**Fixed in this branch (post-review):**
- SSRF integer/hex/octal/short-dotted IP forms rejected (Pydantic + Zod mirror)
- Discovery fingerprint + workspace config hash include literal env/header values (shared `discovery_affecting_payload`)
- Bulk-import secrets in `args` auto-vaulted and resolved sandbox-side
- Discovery schema-cache GC: stale `config_hash` rows purged on upsert; server deletion purges its snapshots
- Proactive apply debounced (1.5s settle) so mutation bursts coalesce into one sandbox apply
- Frontend: memoized server rows + stable handlers, deferred import-parse/modal validation
- Vault secret create/update/delete now invalidates MCP caches when the secret is referenced by a workspace server (version bump + snapshot purge for secret-using servers + proactive apply) — closes the `needs_secret` dead-end and the stale-token discovery snapshot
- Stale-result guard in `discover_and_cache`: a late in-flight discovery re-checks the server's CURRENT fingerprint and drops results for servers edited/deleted mid-discovery (no clobbering a newer snapshot)
- `upsert_tool_schemas` never downgrades a same-hash `ok` row: transient discovery failures annotate (`error` text) instead of wiping cached tools
- Manual `/discover` probe now refreshes the live session composite on an `ok` result (`refresh_session_mcp`), so the agent sees the tools the UI shows without a version bump
- Discovery client clobber race (found via live bulk-import test): concurrent discoveries shared `tools/mcp_client.py`, so a bulk import's parallel probes + background kick overwrote each other's server set and cached spurious "unknown server" errors; each discovery now uploads a unique per-call client under `_internal/` (also keeps probes from replacing the runtime client)

**Deferred (documented, follow-up):**
- No TTL-based re-discovery: an `npx pkg@latest` server keeps the same config hash, so its cached tool snapshot can drift from upstream until a config edit or manual Discover probe (documented in the threat model; `discovered_at` column exists for a future TTL)
- Live-apply race: `proactively_apply_mcp_config` re-syncs a live sandbox immediately rather than strictly on next acquire (deliberate UX trade; race window documented)
- Discovery error isolation hardening, migration enum CHECK constraints, command-allowlist threat-model note

## Summary by CodeRabbit

* **New Features**
  * Full workspace MCP management: add/edit/delete servers, bulk JSON import, promote to templates, test/discover, discovery results view, templates management, lifecycle/status UI, and secrets picker/inline-create.
  * User MCP catalog API & UI for personal templates; add-to-workspace flow.

* **Improvements**
  * Stronger sanitization/SSRF/vault handling, safer generated clients, sandbox isolation, deterministic caching, optimistic toggles, polling stability, and discovery caching/invalidation for secrets.

### DIFF
--- a/Dockerfile.sandbox
+++ b/Dockerfile.sandbox
@@ -30,9 +30,10 @@ RUN apt-get update \
         curl ripgrep jq git unzip \
         libreoffice gcc poppler-utils pandoc qpdf \
         fonts-noto-cjk \
-    # -- uv (Python package manager) --
+    # -- uv (Python package manager) — relocate BOTH uv and uvx; the MCP
+    #    command allowlist permits `uvx`, so it must be on PATH too. --
     && curl -LsSf https://astral.sh/uv/install.sh | sh \
-    && mv /root/.local/bin/uv /usr/local/bin/uv \
+    && mv /root/.local/bin/uv /root/.local/bin/uvx /usr/local/bin/ \
     # -- Node.js (direct binary — avoids apt mirror flakiness) --
     && NODE_ARCH=$([ "$(dpkg --print-architecture)" = "arm64" ] && echo "arm64" || echo "x64") \
     && curl -fsSL "https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-${NODE_ARCH}.tar.xz" \
--- a/migrations/versions/012_add_mcp_servers.py
+++ b/migrations/versions/012_add_mcp_servers.py
@@ -1,0 +1,121 @@
+"""Add per-workspace, user-configured MCP server tables.
+
+Adds the catalog of user-level MCP server templates (user_mcp_servers),
+the per-workspace source of truth (workspace_mcp_servers), the discovery
+schema cache keyed by config version (workspace_mcp_tool_schemas), and a
+mcp_config_version counter on workspaces that every workspace-row mutation
+bumps so sessions can detect config drift without a new per-turn query.
+
+Secrets are never stored here — env/header values hold "${vault:NAME}"
+references resolved against workspace_vault_secrets at sandbox runtime.
+
+Revision ID: 012
+Revises: 011
+"""
+
+from alembic import op
+
+
+revision = "012"
+down_revision = "011"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # User-level templates: UI convenience, copied into a workspace on
+    # "add to workspace" — never inherited at runtime.
+    op.execute("""
+        CREATE TABLE IF NOT EXISTS user_mcp_servers (
+            user_mcp_server_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+            user_id VARCHAR(255) NOT NULL,
+            name VARCHAR(255) NOT NULL,
+            transport VARCHAR(16) NOT NULL DEFAULT 'stdio',
+            command TEXT NULL,
+            args JSONB NOT NULL DEFAULT '[]',
+            url TEXT NULL,
+            env JSONB NOT NULL DEFAULT '{}',
+            headers JSONB NOT NULL DEFAULT '{}',
+            description TEXT NOT NULL DEFAULT '',
+            instruction TEXT NOT NULL DEFAULT '',
+            tool_exposure_mode VARCHAR(16) NOT NULL DEFAULT 'summary',
+            created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+            updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+            UNIQUE(user_id, name)
+        )
+    """)
+
+    op.execute("""
+        CREATE INDEX IF NOT EXISTS idx_user_mcp_servers_user
+        ON user_mcp_servers(user_id)
+    """)
+
+    op.execute("DROP TRIGGER IF EXISTS update_user_mcp_servers_updated_at ON user_mcp_servers")
+    op.execute("""
+        CREATE TRIGGER update_user_mcp_servers_updated_at
+        BEFORE UPDATE ON user_mcp_servers
+        FOR EACH ROW EXECUTE FUNCTION update_updated_at_column()
+    """)
+
+    # Per-workspace source of truth: one row per active user server, plus one
+    # disable-marker row per disabled built-in.
+    op.execute("""
+        CREATE TABLE IF NOT EXISTS workspace_mcp_servers (
+            workspace_mcp_server_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+            workspace_id UUID NOT NULL REFERENCES workspaces(workspace_id) ON DELETE CASCADE,
+            name VARCHAR(255) NOT NULL,
+            source VARCHAR(16) NOT NULL,
+            enabled BOOLEAN NOT NULL,
+            config JSONB NULL,
+            created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+            updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+            UNIQUE(workspace_id, name)
+        )
+    """)
+
+    op.execute("""
+        CREATE INDEX IF NOT EXISTS idx_workspace_mcp_servers_workspace
+        ON workspace_mcp_servers(workspace_id)
+    """)
+
+    op.execute("DROP TRIGGER IF EXISTS update_workspace_mcp_servers_updated_at ON workspace_mcp_servers")
+    op.execute("""
+        CREATE TRIGGER update_workspace_mcp_servers_updated_at
+        BEFORE UPDATE ON workspace_mcp_servers
+        FOR EACH ROW EXECUTE FUNCTION update_updated_at_column()
+    """)
+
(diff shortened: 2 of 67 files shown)


Chen-zexi:
Verified all 7 findings from the [latest review](https://github.com/ginlix-ai/LangAlpha/pull/263#issuecomment-4687474551). 1 of 2 blockers is real; fixed in b7d4ad57 + tests in ea0302a7.

**1. `get_tool_schemas(session.conversation_id)` — refuted.** `Session.conversation_id` *is* the workspace UUID: `WorkspaceManager` calls `SessionManager.get_session(workspace_id, core_config)`, whose first parameter is `conversation_id` and is stored verbatim (`src/ptc_agent/core/session.py:377-389`). The cache read therefore filters `workspace_mcp_tool_schemas.workspace_id` with the correct value. The name is misleading (pre-existing), but the behavior is correct.

**2. `param_type` embedded verbatim in codegen — valid, fixed (b7d4ad57).** Confirmed: discovery sanitized only the tool description, not `input_schema`, so a hostile `properties.*.type` reached both the generated docstring (`_generate_docstring`) and the markdown docs (`generate_tool_documentation`) unescaped. `param_type` now rides the same source-conditional escape as descriptions in both sites. While fixing, also found that a non-str `type` (e.g. a dict) crashed the closed `type_map`/`examples` lookups with `TypeError: unhashable` — both now coerce to safe fallbacks. New tests pin all three (triple-quote breakout inert, unhashable type survives, docs path sanitized) plus builtin byte-stability; all three injection tests fail on the pre-fix code.

**3. Non-atomic purge + version bump in `_invalidate_mcp_for_secret` — valid, fixed (b7d4ad57).** Added `delete_tool_schemas_and_bump(workspace_id, server_names)` which deletes via `server_name = ANY(%s)` and bumps `mcp_config_version` in one transaction; vault invalidation now uses it (falling back to the plain bump when nothing needs purging). Existing invalidation tests updated to the atomic contract.

**4. `_builtin_names()` startup race — refuted.** `agent_config` is assigned in `app_lifespan` (`setup.py:217`) before the `yield`; FastAPI completes the lifespan before accepting traffic, so no request can hit the collision guard while it is `None`. The empty-set fallback is defensive, not reachable.

**5. `promote_server` name divergence — refuted.** `workspace_row_to_server_config` forces `config["name"] = row["name"]` (`mcp_config.py:62`) and every workspace-row write goes through the validated `to_config_blob()`, so the blob name cannot diverge from the row name. Even hypothetically, the catalog upsert is keyed by `(user_id, name)`, so it would not touch a different user's or unrelated row.

**6. `wm._sessions` private access — acknowledged, not changing.** This matches the existing metadata-only read pattern at `threads.py`, `workspace_files.py`, and `secret_redactor.py`; adding a public accessor for one of four call sites would be inconsistent. A broader encapsulation refactor is out of scope for this PR.

**7. Serial upserts in `discover_and_cache` — acknowledged, not changing.** Discovery itself (the up-to-30s part) already runs as a single parallel batch via `discover(servers)`; the serial upserts add roughly a second worst-case at the 20-server cap. A `gather` would grab up to 20 pool connections from a background task and contend with the request path — the serial loop is deliberate backpressure.

Chen-zexi:
Round-4 verdicts: 1 fixed, 1 refuted, 2 skipped with reasons. All non-blocking.

**1. `_get_live_sandbox` reaches into `wm._sessions` — skipped (re-adjudicated in round 2).**
Same finding as round 2's #6. The metadata-only `_sessions.get()` read is the established pattern across `workspace_files.py` and `secret_redactor.py` as well; adding a public accessor used by one of three call sites would be a half-migration. If we introduce `get_session_if_ready()`, it should be a dedicated cleanup that converts all three peers — out of scope for this PR.

**2. Module-level proactive-apply task sets — skipped.**
Accurate observation, but it is the file-scoped convention here: `threads.py` (`_background_tasks`), `workspaces.py` (`_warm_tasks`), and `handlers/chat/ptc_workflow.py` (`_background_tasks`) all hold fire-and-forget task refs at module level. The cross-test leak doesn't materialize in the suite: the only test that exercises the path (`test_proactive_apply_coalesces_burst_mutations`) awaits its tasks inside its own event loop and asserts `_proactive_apply_pending` has drained before returning, and the done-callback discards refs from both containers. A reset fixture would be guarding a state that's already empty at test exit.

**3. Duplicated header branch in `_format_server_brief` / `_format_server_detailed` — fixed in f6f5ecb0.**
Confirmed byte-identical 14-line block. Extracted to `_server_header_lines(server_name, config)`; both formatters now call it and diverge only in their tool-listing tails. Output is byte-identical — the existing neutral-framing and summary-determinism tests pin it, and the full suite (4,135) passes unchanged.

**4. `update_catalog_server` "lacks a runtime guard" — refuted.**
The guard exists: the update loop only appends a column when `col in _jsonb_cols` or `col in _scalar_cols` (`src/server/database/mcp_servers.py:157-163`); any other key is dropped before the f-string is built, so an attacker-controlled map would be filtered, not interpolated. Both call sites additionally pass hardcoded literal keys sourced from the Pydantic model. The allowlist is declared three lines above the loop with a `# Whitelist mutable columns` comment, which serves the documentation purpose the suggested docstring warning would.

PR #265, opened 2026-06-13
feat(web): chat streaming lifecycle + navigation polish (+ subagent reasoning separators)

Chen-zexi:
## Summary

Frontend polish across the chat streaming UI and in-chat navigation, plus a small
backend fix for how subagent reasoning streams. Three areas:

**Streaming live→accordion lifecycle** (`MessageList.tsx`, `ActivityBlock.tsx/.css`)
- Completed tool/reasoning rows now fold into the accordion **immediately on stream
  end** instead of lingering up to 5s. Cooldown shortened (5000ms → 1800ms) so
  sub-second tool calls still register without flicker.
- Smoother handoff: spring-based accordion container + per-row entrance, tightened
  live-row exit, icon pulse synced to the shimmer, completing-state dim + settle.
- Replay/reconnect parity: history threads (`isStreaming=false`) render pre-folded
  with no entrance animation.

**Navigation** (`NavigationPanel.tsx`, `useNavigationData.ts`, `ChatView.tsx`,
`components/nav/`, `WorkspaceGallery.tsx`)
- Workspace **rename** (inline edit) wired into both the nav pill options menu and the
  gallery card menu, with optimistic update + cache invalidation (`lists()` + `detail(id)`).
- **Pin-to-top**, a **new-thread** button (opens a fresh thread in that workspace),
  an **options** (⋯) menu, and **display options** on the nav pill.
- **Stable thread ordering** — the active thread no longer jumps to the top mid-
  conversation; recency order refreshes on workspace switch / reload.
- Unified rail + `BottomTabBar` nav config (`components/nav/navItems.ts`,
  `useNavActive.ts`) with Radix tooltips on the icon rail.
- Removed the AccountMenu theme switcher (theme still lives in Settings).
- Mobile top bar: swapped back-arrow and hamburger order.

**Console cleanup** (`useChatMessages.ts`, `streamEventHandlers.ts`, `useCardState.ts`,
`ChatView.tsx`)
- Stripped the high-frequency `[History]` / `[Stream]` / subagent / card-update debug
  logs that flooded the console during streaming and history replay (one log fired
  16k+ times on a large thread). Kept all `console.warn` / `console.error` diagnostics.

**Backend — subagent reasoning section separators** (`subagent.py`, `content_utils.py`,
`streaming_handler.py`)
- The subagent token forwarder now tracks the OpenAI `summary_text` index and prepends
  a blank line when it changes (0→1), so a new reasoning section's bold title no longer
  glues onto the previous section's prose. Mirrors the main-agent streaming handler.

## Diff Size
- Total: 35 files changed, +3343 / −578
- Net (excl. tests): 29 files changed, +1948 / −577

## Test Coverage
- Frontend: **1544 tests pass** (135 files), tsc clean, `vite build` succeeds.
  New test files: `MessageList.lifecycle.test.tsx`, `ActivityBlock.lifecycle.test.tsx`,
  `useNavigationData.test.tsx`, `nav/useNavActive.test.tsx`, plus expanded
  `NavigationPanel.test.tsx` (lifecycle fold timing, stable ordering with optimistic
  pin/rename rollback, nav active-state matching, workspace controls).
- Backend: **20 forwarder tests pass** (`test_subagent_token_forwarder.py`), ruff clean —
  cover the summary-index separator (inserts on transition, skips within a section),
  reasoning lifecycle signals, and message-id reset.

## Pre-Landing Review
Two parallel adversarial review passes (frontend + backend):
- **Backend:** no findings. Verified the separator-pending flag can't get stuck, the
  top-level vs `summary_text` `index` fields are correctly distinguished, and the moved
  `msg_id` assignment is safe.
- **Frontend:** P0 was a staging issue (new source/test files untracked) — fixed by
  committing them. Two P2 (informational): the `vite.config.js` `input` guard — reconciled
  in the `origin/main` merge and verified by a clean production build; a low-probability
  rename-draft clobber in `WorkspaceGallery` if a background refetch returns a *different*
  name mid-edit (same-value refetch won't trigger it).

## Test plan
- [x] Frontend: 1544 Vitest tests pass; tsc clean; `vite build` succeeds
- [x] Backend: 20 subagent-forwarder tests pass; ruff clean
- [x] Merged latest `main` (incl. #264 onboarding); conflict in `vite.config.js` resolved + build-verified

## Summary by CodeRabbit

* **New Features**
  * Desktop nav pin/unpin, persistent pin state, drag‑and‑drop workspace reordering, inline workspace rename, navigation display options (workspace limit, thread page size, sort modes), tooltips on nav items, workspace/thread paging and “load more”.

* **Bug Fixes**
  * Streaming message formatting: reasoning summary sections now include proper separators to prevent header/prose concatenation.

* **Behavior Changes**
  * Chat folding/live‑row timing adjusted to fold completed items sooner when streams end.

* **Style**
  * Refined animations, badges, sidebar accents and layout tweaks.

* **Tests & Localization**
  * Expanded test coverage and added English/Chinese UI strings.

### DIFF
--- a/src/llms/content_utils.py
+++ b/src/llms/content_utils.py
@@ -213,6 +213,32 @@ def _extract_text_from_summary(summary: Any) -> Optional[str]:
     return combined_text if combined_text else None
 
 
+def extract_reasoning_summary_index(content: Any) -> Optional[int]:
+    """Extract the summary_text item index from reasoning content.
+
+    OpenAI Response API streams reasoning chunks where each ``summary_text``
+    item carries an ``index`` (0, 1, 2…) identifying its reasoning "thought
+    step". When that index changes across chunks (e.g. 0→1) a new reasoning
+    section has started, so streaming callers insert a blank-line separator
+    before it — otherwise the next section's ``**Title**`` glues onto the end
+    of the previous section's prose. Returns None when the content is not
+    reasoning or carries no summary index.
+
+    Note: the top-level reasoning dict also has an ``index`` field, but that is
+    the position in the content array (always 0) — NOT the thought-step index.
+    """
+    items = [content] if isinstance(content, dict) else (content if isinstance(content, list) else [])
+    for item in items:
+        if not isinstance(item, dict) or item.get("type") != "reasoning":
+            continue
+        summary = item.get("summary")
+        if isinstance(summary, list):
+            for s in summary:
+                if isinstance(s, dict) and "index" in s:
+                    return s["index"]
+    return None
+
+
 # ============================================================================
 # LangChain Message Content Extraction
 # ============================================================================
--- a/src/ptc_agent/agent/middleware/background_subagent/subagent.py
+++ b/src/ptc_agent/agent/middleware/background_subagent/subagent.py
@@ -26,6 +26,7 @@ from ptc_agent.agent.middleware.background_subagent.middleware import (
 )
 from ptc_agent.agent.middleware.background_subagent.registry import BackgroundTaskRegistry
 from ptc_agent.agent.middleware._utils import append_to_system_message
+from src.llms.content_utils import extract_reasoning_summary_index
 from src.llms.llm import narrow_prompt_cache_key
 from src.server.utils.content_normalizer import normalize_text_content
 
@@ -65,6 +66,12 @@ class _SubagentTokenForwarder:
         self.agent_id = agent_id
         self._reasoning_active = False
         self._last_msg_id: str | None = None
+        # Track the OpenAI reasoning summary_text index to separate sections.
+        # When it changes (0→1) a new reasoning section starts; we prepend a
+        # blank line so its `**Title**` header doesn't glue onto the previous
+        # section's prose. Mirrors WorkflowStreamHandler's main-agent path.
+        self._reasoning_block_index: int | None = None
+        self._reasoning_separator_pending = False
 
     def _signal_record(self, msg_id: str, content: str) -> dict[str, Any]:
         return {
@@ -133,6 +140,30 @@ class _SubagentTokenForwarder:
         if metadata is not None and metadata.get("langgraph_node") == "tools":
             return
 
+        msg_id = message_chunk.id or f"sg-{self.tool_call_id}"
+
+        # A new assistant message begins a fresh reasoning stream — drop any
+        # carried-over section index / pending separator so the first chunk of
+        # the new message isn't falsely prefixed with a blank line.
+        if self._last_msg_id is not None and msg_id != self._last_msg_id:
+            self._reasoning_block_index = None
+            self._reasoning_separator_pending = False
+
+        # Detect reasoning summary_text index transitions (mirror of the main
+        # streaming handler): when the OpenAI summary index changes (0→1) a new
+        # reasoning section started — queue a separator so its `**Title**` header
+        # lands on its own line instead of gluing onto the previous section's
+        # prose. The flag is pending because the index can arrive before the
+        # chunk that carries the new section's first emittable text.
+        reasoning_idx = extract_reasoning_summary_index(message_chunk.content)
+        if reasoning_idx is not None:
+            if (
+                self._reasoning_block_index is not None
+                and reasoning_idx != self._reasoning_block_index
+            ):
+                self._reasoning_separator_pending = True
+            self._reasoning_block_index = reasoning_idx
+
         # Reasoning content can ride on either ``content`` or
         # ``additional_kwargs.reasoning[_content]`` depending on provider.
         text, content_type = normalize_text_content(message_chunk.content)
@@ -145,8 +176,6 @@ class _SubagentTokenForwarder:
             if r_text:
                 text = r_text
                 content_type = "reasoning"
-
-        msg_id = message_chunk.id or f"sg-{self.tool_call_id}"
 
         # New message id with reasoning still active → close out the old one.
         if (
@@ -172,6 +201,14 @@ class _SubagentTokenForwarder:
                     self.tool_call_id, self._signal_record(msg_id, "complete")
                 )
                 self._reasoning_active = False
+                # Reasoning ended — reset section tracking for the next stream.
+                self._reasoning_block_index = None
+                self._reasoning_separator_pending = False
+
+            # Prepend the blank-line separator queued by a section transition.
+            if content_type == "reasoning" and self._reasoning_separator_pending:
+                text = "\n\n" + text
+                self._reasoning_separator_pending = False
 
             await self.registry.append_captured_event(
                 self.tool_call_id,
--- a/src/server/handlers/streaming_handler.py
+++ b/src/server/handlers/streaming_handler.py
@@ -21,6 +21,7 @@ from src.server.utils.content_normalizer import (
     normalize_text_content,
     is_thinking_status_signal,
 )
+from src.llms.content_utils import extract_reasoning_summary_index
 from src.utils.tracking import ExecutionTracker
 from src.config.settings import (
     get_workflow_timeout,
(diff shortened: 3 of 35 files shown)


Chen-zexi:
## Review response

Verified each finding against current code and pushed `9c1cd7c` (fixes) + `a3f7fb2` (test). Applied 6, skipped 2 Majors + the alias nits with reasons below.

### Applied
- **AccountMenu separator** — gated the divider on `accountUrl`; OSS mode (`accountUrl === null`) no longer stacks a stray separator against the profile divider.
- **NavDisplayOptions button type** — added `type="button"` to the `ChoiceGroup` buttons.
- **Threads cache key (Major)** — the query key now carries `threadPageSize` (`[...byWorkspace(id), threadPageSize, 0]`, matching the dashboard-widget thread keys), so changing the per-workspace thread count refetches instead of replaying a stale page. Prefix-invalidation from `ChatView`/`ThreadGallery` still matches the extended key; added a regression test.
- **Rename draft clobber (claude review P2)** — the rename effect no longer depends on `workspace.name`; the draft is seeded once on entry, so a background refetch can't overwrite an in-progress edit.
- **HMR port guard** — `VITE_HMR_CLIENT_PORT` is finite-checked; a non-numeric value falls back to default HMR behavior instead of `NaN`.
- **Test import alias** — `../../../../test/utils` → `@/test/utils`.

### Skipped (with evidence)
- **`resetStableNavOrder` reset event (Major)** — `resetStableNavOrder` has no production callers (`grep` across `src/`): it's a test-only reset invoked in `beforeEach` before any view mounts, so there are no subscribers to notify. The listener-notify would be dead code.
- **Rename draft lost on persist failure (Major)** — the component's error model is uniformly optimistic-update + rollback + `console.error`, no toast (pin toggle included); the rollback visibly reverts the name. Reopening the editor on failure conflicts with the `renamingWsIdRef` idempotency guard and would introduce toast infra this surface doesn't use. Tracking as a follow-up rather than a one-off.
- **Remaining `@/` alias nits** (`Sidebar`, `BottomTabBar`, `useNavActive`) — short 1-level sibling imports in files otherwise untouched; converting a same-dir `./navItems` to `@/components/nav/navItems` is strictly worse. The directory is already mixed (52 `@/` vs 77 relative).

Green after the changes: 1545 frontend tests, `tsc`, `eslint`, `vite build`.

Chen-zexi:
## Review response (round 2)

Verified the 3 non-blocking findings against current code; pushed `185fa91` (fixes) + `af14c26` (tests). Applied 2, skipped 1.

### Applied
- **`_expandedWorkspaces` retains deleted IDs → spurious 404 on remount** — fixed at the root: a new `forgetNavPanelExpansion(wsId)` is called from the workspace delete path, so the mount-effect no longer re-expands a deleted workspace. Note the suggested mount-effect filter (`if (byId.has(wsId))`) isn't safe here — that effect is mount-only (`[]` deps) so `workspaces` may still be empty, and `workspaces` is the paged/limited slice, so a scrolled-out (not deleted) workspace would be wrongly dropped. Deletion is the only sound prune signal. Added a remount test.
- **`loadMoreThreads` duplicate fetch on rapid double-tap** — added a per-workspace in-flight `Set` guard (mirrors the existing `wsFetchRef` single-flight). The offset is snapshotted before the `await`, so the second tap now no-ops instead of refetching the same page. Added a double-tap regression test.

### Skipped
- **`patchWorkspace` swallows rename/pin errors silently** — same call as the prior round's rename-on-failure item. The component's error model is uniformly optimistic-update + rollback + `console.warn` with no toast (pin and delete included); the rollback visibly reverts the change. Surfacing a toast is a deliberate cross-cutting follow-up, not a one-off here.

Green: 1547 frontend tests, `tsc`, `eslint`, `vite build`.

Chen-zexi:
## Review response (round 3)

Verified all 7 findings; pushed `886bca3`. Applied 3, deferred 2 (documented), and 2 are deliberate decisions confirmed with the maintainer.

### Applied
- **#3 `any` casts on optimistic rollback** — replaced `[unknown, any]` / `key as any` in `WorkspaceGallery` with the typed `CachedWorkspaceList` + `key as readonly unknown[]` pattern already used by `patchWorkspace`. No `any`, no behavior change.
- **#6 frozen-order accumulation on delete** — added `forgetStableNavOrder(wsId)` and wired it into the workspace delete path (next to the round-2 expansion prune), so `_frozenThreadOrders` / `_frozenWorkspaceOrder` / `_lastWorkspaceArrangement` drop the deleted id instead of retaining a ghost. (`resetStableNavOrder` itself is still test-only by design; the real gap was the delete path, now closed.)
- **#7 minimize button a11y/i18n** — now uses `t('nav.minimize')` (new key, en + zh) for both `title` and `aria-label`, matching the sibling pin button.

### Decisions (confirmed with maintainer)
- **#1 impure `useMemo` (marked blocking)** — the anti-pattern is real, but the concurrent-render race it describes **cannot fire as the app is wired**: there is no `StrictMode` and no `useTransition`/`Suspense` on the nav path, so this tree renders synchronously and the memo factory runs once per render (no interleave/discard). A correct fix reworks the whole ordering subsystem (this memo + `orderThreads` + `reorderWorkspace` share the pattern, 34 tests). Decision: **defer**, with an in-code invariant note documenting the constraint and the rework trigger (if concurrent features are ever added here). Latent-only, zero present-day misbehavior.
- **#5 `workspaceLimit` default `'all'`** — intentional. The nav reads as a browsable workspace tree; only workspace **names** are paged (100 at a time beyond the first 20) and threads stay lazy per expand, so first-load is metadata-only. Added a comment recording the intent.

### Deferred (non-blocking, with reasons)
- **#2 stale offset in `loadMoreThreads`** — low severity: the success handler dedups by id (no duplicate rows) and the round-2 single-flight guard covers concurrent user taps. The proposed "read offset inside the updater" can't cleanly issue the async fetch from a setState updater. Worst case is one short page on a click.
- **#4 `_reasoning_separator_pending` text-branch reset** — as noted in the finding, this can't occur on the normal OpenAI streaming path; touching the well-tested `subagent.py` for a non-occurring edge isn't worth reopening the backend slice here.

Green: 1547 frontend tests, `tsc`, `eslint`, `vite build`.

Chen-zexi:
## Review response (round 4)

The single non-blocking finding (reorder sends full records instead of projected pairs) is a **false positive** — the projection is already in place at both call sites, so there's no payload bloat and no type mismatch. No change needed.

`WorkspaceGallery.tsx` builds `items` as projected pairs before the call:

```ts
// WorkspaceGallery.tsx:739-743
const items: { workspace_id: string; sort_order: number }[] = [];
reordered.forEach((ws, i) => {
  if (ws.status === 'flash') return;
  items.push({ workspace_id: ws.workspace_id, sort_order: i });
});
// :754
await reorderWorkspaces(items);   // already { workspace_id, sort_order }[], not WorkspaceRecord[]
```

The nav path projects identically:

```ts
// useNavigationData.ts:285-287
.map((ws, i) => ({ workspace_id: ws.workspace_id, sort_order: i }));
await reorderWorkspaces(items);
```

`items` matches `reorderWorkspaces(items: Array<{ workspace_id: string; sort_order: number }>)` (api.ts:87) exactly, so `tsc` is clean — no `unknown`/`WorkspaceRecord[]` widening. The `position` → `sort_order` rename you noted as correct is the only field change; the projection itself was never dropped.

On the rest: the impure-`useMemo` comment stays as the documented tripwire before StrictMode/`useTransition` lands on that path.

Chen-zexi:
## Review disposition — summary across all rounds

All automated review feedback (four passes) is now dispositioned. Consolidated view for human reviewers:

### Applied (`9c1cd7c`, `185fa91`, `886bca3` + tests `a3f7fb2`, `af14c26`)
- **AccountMenu** — gate the divider on `accountUrl` so OSS mode (`accountUrl === null`) doesn't stack a stray separator.
- **NavDisplayOptions** — explicit `type="button"` on the `ChoiceGroup` buttons.
- **Threads cache key** — keyed by `threadPageSize` (`[...byWorkspace(id), threadPageSize, 0]`) so changing the per-workspace count refetches instead of replaying a stale page. _(regression test)_
- **Rename draft** — seeded once on entry; a background refetch no longer clobbers an in-progress edit.
- **HMR port guard** — `VITE_HMR_CLIENT_PORT` finite-checked (no `NaN`).
- **Nav expansion prune** — `forgetNavPanelExpansion` on delete stops a remount re-expanding a deleted workspace (no spurious 404). _(regression test)_
- **Single-flight `loadMoreThreads`** — per-workspace in-flight guard kills the rapid double-tap double-fetch. _(regression test)_
- **Typed rollback casts** — dropped `any` in `WorkspaceGallery` for the `CachedWorkspaceList` pattern.
- **Frozen-order prune** — `forgetStableNavOrder` on delete clears ghost ordering entries.
- **Minimize button** — i18n `title` + `aria-label` (`nav.minimize`, en + zh).

### Decisions documented in-code
- **Impure `useMemo`** (marked blocking) — deferred: no `StrictMode`/`useTransition`/Suspense on the nav path, so the concurrent-render race can't fire today. Added an invariant comment naming the rework trigger; a follow-up should make the ordering subsystem pure if concurrent features land there.
- **`workspaceLimit: 'all'` default** — intentional (names-only paging, threads stay lazy); intent comment added.

### Deferred non-blocking (with reasons)
- **Stale `loadMoreThreads` offset on background refetch** — deduped by id + single-flight already cover the realistic cases; worst case is one short page.
- **Backend `_reasoning_separator_pending` text-branch reset** — can't occur on the normal OpenAI streaming path.

### False positive
- **Reorder "sends full records"** — both call sites already project to `{ workspace_id, sort_order }` (`WorkspaceGallery.tsx:739-743`, `useNavigationData.ts:285`); `tsc` clean, signature matches exactly.

Gate green: **1547** frontend tests, **20** backend tests, `tsc`, `eslint`, `vite build`.

PR #272, opened 2026-06-14
fix(server): atomic JSONB append for context_window SSE persistence

Chen-zexi:
## Summary

Replaces the read-modify-write that persisted `context_window` SSE events with an atomic server-side JSONB append, closing a race that silently dropped events on replay.

`_persist_context_window_event` (fired by manual `/compact` and `/offload`) used to load *every* response for the thread via `get_responses_for_thread`, append in Python, and rewrite the whole `sse_events` blob with `update_sse_events`. Under concurrency that clobbered overlapping appends — last writer wins. New `append_sse_event` does a single `UPDATE ... SET sse_events = COALESCE(sse_events,'[]'::jsonb) || %s::jsonb` scoped to the latest response by `turn_index`, which is race-free under the autocommit pool (row-lock + re-evaluation).

**Backend** (`conversation.py`, `workflow_handler.py`)
- Add `append_sse_event(thread_id, event, conn=None)` — atomic JSONB concat to the thread's most-recent response. Mirrors `update_sse_events`' conn/pool pattern and the NUL-safe `SafeJson` bind.
- Route `_persist_context_window_event` through it; stays best-effort (logs, never raises). Latest-response semantics preserved (old `responses[-1]` over `turn_index ASC` == new `ORDER BY turn_index DESC LIMIT 1`).

**Tests**
- Unit: SQL shape (`||`, `ORDER BY turn_index DESC LIMIT 1`), one-element JSONB array bind, no-row → False, pool vs passed-conn paths, handler routing + best-effort swallow.

## Diff Size
- Total: 2 files changed, 59 insertions(+), 19 deletions(-)
- Net (excl. tests): same (+2 untracked test files added)

## Test Coverage

`append_sse_event`: conn path, pool path, row-hit (True), no-row (False). `_persist_context_window_event`: routes to append, best-effort on failure.

Verified end-to-end against real Postgres (not mocks): sequential ordering, latest-response targeting, `COALESCE(NULL)`, no-row case. Plus a 25-way concurrency test — **new code keeps 25/25 events; the old read-modify-write kept 1/25** under identical load. The race was real and is now fixed.

## Pre-Landing Review

Clean — 0 critical, 1 informational: a pre-existing append-vs-full-overwrite interaction with the SubagentCollector (`background_task_manager.py`), unchanged by this PR and strictly improved (the old path also lost appends among themselves).

## Test plan
- [x] 5 new unit tests pass
- [x] 461 surrounding backend unit tests pass; ruff clean
- [x] Real-Postgres e2e: targeting / ordering / COALESCE / no-row + race-free (25/25)

## Summary by CodeRabbit

* **Improvements**
  * Improved reliability of real-time event streaming by atomically appending SSE events to the most recent conversation response
  * Improved robustness of workflow context window event persistence using best-effort append behavior to avoid concurrent update conflicts

* **Tests**
  * Added unit tests covering atomic JSONB append behavior, correct return values when no row is found, connection reuse behavior, and best-effort failure handling for context window persistence

### DIFF
--- a/src/server/database/conversation.py
+++ b/src/server/database/conversation.py
@@ -1348,6 +1348,54 @@ async def update_sse_events(
         raise
 
 
+async def append_sse_event(
+    conversation_thread_id: str,
+    event: Dict[str, Any],
+    conn=None,
+) -> bool:
+    """Atomically append one SSE event to the thread's latest response blob.
+
+    A server-side ``sse_events || event`` JSONB concat scoped to the most-recent
+    response (by turn_index). Avoids reading the whole blob into Python and
+    rewriting it, and is race-free against concurrent appenders (the
+    read-modify-write it replaces is not). Returns True when a row was updated,
+    False when the thread has no response row yet.
+    """
+    sql = """
+        UPDATE conversation_responses
+        SET sse_events = COALESCE(sse_events, '[]'::jsonb) || %s::jsonb
+        WHERE conversation_response_id = (
+            SELECT conversation_response_id
+            FROM conversation_responses
+            WHERE conversation_thread_id = %s
+            ORDER BY turn_index DESC
+            LIMIT 1
+        )
+    """
+    params = (SafeJson([event]), conversation_thread_id)
+    try:
+        if conn:
+            async with conn.cursor() as cur:
+                await cur.execute(sql, params)
+                updated = cur.rowcount > 0
+        else:
+            async with get_db_connection() as conn:
+                async with conn.cursor() as cur:
+                    await cur.execute(sql, params)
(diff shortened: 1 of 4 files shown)


Chen-zexi:
Removed the redundant no-row DEBUG line in `append_sse_event` (1930bfee). The helper now returns `False` on the no-row case without logging; the caller's `_persist_context_window_event` already emits the contextual "...skipping context_window persist" line, so that path logs exactly once now.

The other points (turn_index targeting, JSONB concat / NULL handling, autocommit row-lock serialization, best-effort swallow, call-time import patching) were verified correct — no changes there.

PR #280, opened 2026-06-22
feat(agent): adopt deepagents 0.6 DeltaChannel for messages checkpoints

lastfreeyang-lgtm:
## 배경
deepagents v0.6 + LangGraph 1.2에서 도입된 **DeltaChannel**(스텝당 messages 전체 스냅샷 대신 증분 diff만 체크포인트)을 메인 에이전트에 적용합니다. 장기 분석 세션에서 Postgres checkpointer 쓰기 부하와 재연결 시 state 로딩 비용을 크게 줄입니다.

## 변경
1. `deepagents` `0.5.3 → 0.6.4+`, `langgraph` `0.3.5 → 1.2+` 범프 (lock 동반)
2. 메인 `create_agent(...)`에 `state_schema=DeepAgentState` 전달

## 중요 포인트 (자동 적용 아님)
deepagents 0.6은 **`create_deep_agent` 경로에만** DeltaChannel을 기본 적용합니다. 우리처럼 `langchain.agents.create_agent`로 만든 에이전트는 명시적으로 DeltaChannel-annotated `state_schema`를 넘기지 않으면 messages가 그대로 `add_messages`(`BinaryOperatorAggregate`, 풀 스냅샷)로 남습니다 — 그래서 `state_schema=DeepAgentState`를 명시했습니다.
- `DeepAgentState`는 deepagents **0.6.4**부터 public (0.6.3은 private `_DeepAgentState`만 존재) → pin을 `>=0.6.4`로 설정.

## 측정 (실제 AsyncPostgresSaver)
40스텝 분석세션(큰 AI + tool 메시지) 기준:

| 항목 | 결과 |
|---|---|
| `checkpoint_blobs` 저장량 | **395KB → 12KB (~32×↓)** |
| 재연결(state replay) 후 메시지 내용/순서 | 완전 보존 ✅ |
| 재연결 후 이어서 델타 append | 정상 ✅ |
| compaction(RemoveMessage 전체 덮어쓰기) 호환 | 정상 ✅ |
| 유닛테스트 | 4572 passed (무회귀) |

## 비고
- `files` 채널은 자체 `FileOperationMiddleware`라 여전히 스냅샷 — 절감 본체는 messages라 영향 작으며, 후속으로 델타화 가능.
- `snapshot_frequency=50`: 50스텝마다 풀 스냅샷 1회(deepagents 기본값).

## Summary by CodeRabbit

* **Chores**
  * Updated LangGraph dependency (0.3.5 → 1.2.0) and DeepAgents (0.5.3 → 0.6.4).

* **Improvements**
  * Optimized state persistence for long-running sessions. The agent now uses incremental state updates instead of storing full snapshots, improving efficiency and reducing storage overhead during extended operations.


lastfreeyang-lgtm:
## Re: backward-compat of pre-upgrade Postgres checkpoints (blocking) + `snapshot_frequency` (non-blocking)

Thanks for the careful review. Both points addressed below.

### Blocking — old `add_messages` checkpoints resume cleanly under `DeltaChannel`

This is **explicitly handled by langgraph's `DeltaChannel`** (acceptance path #1: documented backward compat). `DeltaChannel.from_checkpoint` recognizes three blob shapes, and the old-format case is called out by name:

```python
# langgraph/channels/delta.py  (DeltaChannel.from_checkpoint)
"""Blob types:
  * MISSING: start empty; caller replays writes.
  * _DeltaSnapshot(value): restore value directly from snapshot.
  * plain value (migration from old BinaryOperatorAggregate blobs): use directly.
"""
...
if checkpoint is MISSING:
    new.value = self.typ()
elif isinstance(checkpoint, _DeltaSnapshot):
    new.value = checkpoint.value
else:                       # <-- old add_messages blob = plain list
    new.value = checkpoint
```

Pre-upgrade threads stored `messages` as a full list via `BinaryOperatorAggregate`. On resume that plain list falls into the `else` branch and is used directly as the channel's base value; subsequent writes proceed through the `DeltaChannel` path normally. No `_DeltaSnapshot` wrapper is required for the old blob, so there is no corruption or read failure on the old→new transition.

The reducer also tolerates `state=None` (threads whose earliest checkpoint never seeded `messages: []`) — see the `state or []` guard in `_messages_delta_reducer`.

**Empirically verified** on the pinned floor range (langgraph 1.2.6, deepagents 0.6.11):

```python
old_blob = [HumanMessage("hi", id="m1"), AIMessage("hello", id="m2")]  # old full-list blob
ch = DeltaChannel(_messages_delta_reducer, list, snapshot_frequency=50); ch.key = "messages"
restored = ch.from_checkpoint(old_blob)
# -> [('m1','hi'), ('m2','hello')]            # old history intact
restored.update([[AIMessage("continued", id="m3")]])
# -> [('m1','hi'), ('m2','hello'), ('m3','continued')]   # new delta write appends correctly
```

So an in-flight thread created under the old schema resumes with its full history and continues writing as deltas from that point. No deploy-time epoch-fencing migration is needed.

### Non-blocking — `snapshot_frequency`

Agreed it's currently implicit (it's the `DeepAgentState` default of 50, set in `deepagents.graph`). We're intentionally keeping the library default for now rather than adding a knob we have no tuning signal for — 50 bounds replay depth well within the system-wide `DELTA_MAX_SUPERSTEPS_SINCE_SNAPSHOT` (5000) ceiling. If/when we want to tune it, the override path is a thin `DeepAgentState` subclass re-annotating `messages` with a different `snapshot_frequency`, which we can wire through `agent_config.yaml` at that point. Happy to add the passthrough now if you'd prefer it surfaced from day one.

Chen-zexi:
Hey @lastfreeyang-lgtm , thanks for the PR. 

I actually have a working branch locally for delta channel support. I will take a deeper look later to see which branch we should keep.

Chen-zexi:
Thanks for the PR and the detailed measurements, @lastfreeyang-lgtm.

We'd been working on DeltaChannel adoption in parallel — #283 lands the same approach with the hardening this branch is missing:

- **All four `create_agent` sites, not just the main agent.** This branch opts in the one call in `agent.py`; #283 also wires Flash mode and the background subagents (two more calls), so those stop writing full per-step snapshots too.
- **Tighter floors.** `langgraph>=1.2.2` (id-stamping via `ensure_message_ids` landed in 1.2.2, not 1.2.0) and `deepagents>=0.6.11` (non-minting reducer + `PatchToolCallsMiddleware` off the id-less `Overwrite`). Below these, reconstruction can drift/duplicate ids on replay.
- **Vendored + parity-tested reducer** instead of importing `DeepAgentState` directly, so an upstream change can't silently alter our on-disk format.
- **Delta-specific tests** (determinism, ordering, resolution, reducer parity, role stamping) + a `_DeltaSnapshot`-aware NUL-scrub fix + the legacy-thread analysis (they reconstruct faithfully — no migration needed).

We'll land #283 and close this one — and we've credited you as co-author on #283 for the overlapping work. Thanks again.

PR #281, opened 2026-06-23
feat(chart-selection): hand chart region/price selections to the agent

Chen-zexi:
## Summary

Adds the **user → agent** direction on the MarketView chart. Previously the agent
could draw annotations onto a chart (`draw_chart_annotation`), but a user had no way
to point at something and ask about it. Now they can.

A user arms a select tool, drag-selects a **region** (time × price box) or clicks a
**price level**, annotates it with a note, and on send the selection rides to the
agent as structured `chart_selection` context — bounds + per-candle OHLCV + the note.
The agent reads that exact region and can draw a `rectangle` / `price_line` back onto
the same `symbol:timeframe`. Cards render on the sent message and replay from turn
metadata on history reload. Works in PTC and Flash, desktop and mobile send paths,
en-US + zh-CN.

**Backend** — new `ChartSelectionContext` (+ typed `SelectionBar`) on the
`additional_context` union; a parse + reminder builder that injects each selection
inline on the user's turn; a shared `inject_inline_reminders` helper used by both the
PTC and Flash workflows; skill-doc guidance for reading a selection and drawing back.

**Frontend** — a select tool + pointer handlers + draft primitive on `MarketChart`; a
`useSyncExternalStore` selection store with an inline note editor and numbered overlay;
a `chartSelectionToContext` mapper; selection cards on the user message; both send
paths share a `buildChartSelectionSend` helper.

## Overview

Net-change map of the whole branch:

| | Files | Lines |
|---|---|---|
| Modified | 21 | +1,112 / −141 |
| New | 15 | +2,463 |
| **Total** | **36** | **+3,575 / −141** |
| **Net (excl. tests)** | **27** | **+2,435 / −141** |

**By module:**

- **Backend — models / utils / handlers** *(new: `chart_selection_context.py`; modified:
  `additional_context.py`, `_common.py`, `ptc_workflow.py`, `flash_workflow.py`)* —
  the `chart_selection` context type, its parse + reminder pipeline, and the shared
  inline-injection helper.
- **Skill** *(`skills/chart-annotation/SKILL.md`)* — "Reacting to a user's chart
  selection" section.
- **Frontend — MarketView** *(new: `chartSelectionStore`, `selectionPrimitive`,
  `downsampleBars`, `selectionSend`, `SelectionCommentOverlay.tsx/.css`; modified:
  `MarketChart.tsx/.css`, `MarketChatPanel.tsx`, `MarketView.tsx`)* — selection
  capture, the store, the overlay/editor, and both send paths.
- **Frontend — ChatAgent** *(new: `SelectionContextPreview.tsx`; modified:
  `fileUpload.ts`, `MessageList.tsx`, `ChatView.tsx`, `useChatMessages.ts`,
  `historyEventHandlers.ts`, `messageHelpers.ts`, `chat-input.tsx`)* — the
  context mapper, selection cards on the user bubble, and history replay.
- **Shared** *(`types/chat.ts`, `locales/en-US.json`, `locales/zh-CN.json`)* —
  `ChartSelectionSnapshot` type + i18n keys.

## Diff Size
- Total: 36 files, +3,575 / −141
- Net (excl. tests): 27 files, +2,435 / −141

## Test Coverage

Every new module has tests. New/extended test files:
- Backend: `test_chart_selection_context.py` (new), `test_additional_context.py`
  (+SelectionBar/geometry), `test_chat_common.py` (+inject_inline_reminders incl.
  HITL-resume / checkpoint-replay no-op guards).
- Frontend: `chartSelectionStore.test.ts`, `selectionSend.test.ts`,
  `downsampleBars.test.ts`, `selectionPrimitive.test.ts`,
  `chartSelectionToContext.test.ts`, `messageHelpers.test.ts` (all new).

## Pre-Landing Review

Ran this session. Findings were fixed, not deferred: extracted the
`inject_inline_reminders` backend helper + a `buildChartSelectionSend` frontend helper
(DRY'd the two send paths), a shared `isConfirmedFor` predicate (4 call sites), typed
`SelectionBar` (NaN/Inf rejection), removed a dead store method, named magic numbers,
and bounded `symbol` length. `tsc` clean; ESLint clean (only pre-existing
`exhaustive-deps` warnings in untouched MarketChart hooks).

## Design Review

Frontend-touching. Lite pass this session: i18n for all new strings (en-US + zh-CN),
aria-labels on the select/ruler/clear toolbar buttons, locale-aware price/volume
formatting on the selection cards.

## Plan Completion

Plan: `chart-selection` user→agent direction — **complete**.
- CHANGED: the cropped-screenshot payload was dropped in favor of structured OHLCV
  bars (the exact bars are the agent's signal; no vision dependency).
- Out of scope (unchanged): clicking an existing agent annotation to reference it,
  server-side persistence of user selections.

## Notes for reviewers

- The selection `label` (user's note) is interpolated verbatim into the inline
  `<system-reminder>`. This is **self-injection** — the user authoring their own
  prompt — so it's low severity (no privilege escalation, no cross-user surface).
  Delimiter hardening was considered and consciously deferred.

## Test plan
- [x] Backend: `uv run pytest tests/unit/server/` — 2005 passed
- [x] Frontend: `pnpm vitest run` — 1899 passed (169 files)
- [x] `tsc --noEmit` clean, `eslint` clean

## Summary by CodeRabbit

* **New Features**
  * Added chart selection support (region + price level) with per-selection notes, chat preview cards, and sendable chip summaries.
  * Users can draw selections on the chart, edit comments via an overlay, and confirm/remove selections for inclusion in outgoing messages.
* **Documentation**
  * Updated SKILL guidance for handling incoming `<chart-selection>` blocks and drawing matching chart annotations.
* **Localization**
  * Added English/Chinese UI strings for selection modes, previews, and chip actions.
* **Bug Fixes**
  * Inline reminder injection now includes chart-selection context, and chart selections are restored correctly during chat history replay.
* **Tests**
  * Expanded unit tests covering selection parsing, validation, rendering, sending, and chat message wiring.


Chen-zexi:
Pushed `ebf46d3d..35647bfa` addressing review feedback. Disposition below — verdicts verified against the actual code, not taken at face value.

## Fixed
- **Missing `pointercancel` cleanup** (MarketChart.tsx) — the capture surface wired only `down`/`move`/`up`, so an interrupted touch or stolen pointer capture left the draft box stuck on screen. Added `onPointerCancel` + `onLostPointerCapture` → idempotent cleanup.
- **OHLCV preview headers not localized** (SelectionContextPreview.tsx) — the only English-only strings in an otherwise bilingual dialog. Now keyed via `marketView.selection.col*` in en-US + zh-CN.
- **`downsampleBars` unguarded `max`** — added a `RangeError` for non-positive / non-integer `max` so the exported util fails loud instead of producing an `Infinity` stride. Covered by a new test.
- **Inline-reminder injection logging** — removed entirely. `inject_inline_reminders` now takes plain reminder strings; the `detail`/`log_prefix` machinery existed only to feed an `info` log that was noise. (This also subsumes the "log fires even when the append no-ops" finding — there's no log left to mislead.)

## Rejected — with evidence
- **"Steering path drops chart selections."** Incorrect. The `[]` at `useChatMessages.ts:3971` is the `messageHistory` positional arg, not the context slot — `additionalContext` (carrying the selections) is forwarded at `useChatMessages.ts:4005`. Selections reach the agent identically on both the normal and steering paths.
- **`@`-alias import nitpicks (×6).** No lint rule enforces `@`-alias here (no import plugin in `eslint.config.js`), and the tree is ~1138 relative vs ~505 alias imports. The new code matches the relative-import style of its sibling files (e.g. `MarketChart.tsx`). Converting these would make the new files *less* consistent with their own directories.
- **`as unknown as Record<string, unknown>[]` double-cast** (selectionSend.ts). This is the established idiom in the same call path — see the identical cast at `MarketChatPanel.tsx:486`. The mapper's return interface is not exported; the destination array is `Record<string, unknown>[]`. Removing the cast would diverge from local convention without net benefit.

## Deferred — intentional / out of scope
- **Escaping user fields in the `<chart-selection>` markup.** Separate scope — there's no precedent in this repo for escaping interpolated context, and the surface is self-injection (the user's own note in their own turn; no cross-user reach). Tracked for a follow-up, not this PR.
- **Overlay price-scale subscription.** Matches the existing `AgentEventOverlay` pattern; lightweight-charts exposes no price-scale-change subscription analogous to the time scale's, so there's no clean hook to add.
- **Selection chips → CSS class** (MarketChatPanel.tsx). The chips deliberately mirror the adjacent `bannerStyle()` inline-style + CSS-var pattern used by the status banners right above them; a class would diverge from that local norm.
- **Double `parse_chart_selection_contexts` call** and **HITL-guard asymmetry (PTC vs Flash).** Both confirmed behavior-equivalent today (HITL resume yields a `Command`, not a dict, so PTC's `isinstance` guard suffices). Cosmetic/defensiveness only; reworking the control flow isn't worth the risk here.

Chen-zexi:
Dispositions on the four non-blocking items, pushed in `534e891a`.

**1. `parse_chart_selection_contexts` called twice — done.** Now parsed once alongside `widget_ctxs` near the top of both workflows and reused for metadata persistence and inline injection; the redundant `selection_ctxs` parse is gone. Matches the widget pattern as suggested.

**2. Logging "regressed" for inline injection — intentional, not re-adding.** The shared `logger.info` line was removed deliberately at the maintainer's request (it was one log covering directive/widget/chart-selection, flagged as noise). The observability tradeoff is acknowledged and accepted; re-adding it would reverse that decision.

**3. Mobile Flash path doesn't forward `selectionSnapshots` — won't fix; the premise doesn't hold for this path.** There is no rendered user bubble on the mobile Flash path to carry selection cards:
- `MarketView.tsx` destructures only `{ isLoading, handleSendMessage }` from `useMarketChat()` — not `messages`. That hook's message state is never rendered (the mobile branch shows the chart + a FAB that only wraps the `ChatInput`; no `MessageList`).
- `useMarketChat`'s `MarketChatMessage` / `createUserMessage` / `handleSendMessage` have no chart-selection field or param.
- Per the in-code comment: *"Mobile FAB still uses the legacy useMarketChat (no persistence). Desktop chat lives in MarketChatPanel."*

The selection **data** already reaches the agent on this path via `imageContext`. Forwarding the snapshots would pass a prop nothing accepts and nothing renders. Card parity on mobile Flash would require building out a message-rendering surface for the legacy FAB path (or routing it through `MarketChatPanel`) — a separate feature, not in scope here. The desktop Flash and mobile PTC paths render cards correctly.

**4. Broaden the self-injection note to `time_start`/`time_end` — agreed.** Confirmed both are interpolated verbatim (40-char bounded, not ISO-validated) into the draw-back hint, same self-injection class as `label`. The escaping/delimiter hardening is deferred to a separate change (no precedent in this repo for escaping interpolated context, and the surface is self-injection only — the user's own note in their own turn). When it lands it will cover `label`, `symbol`, `time_start`, and `time_end`.

Chen-zexi:
Verified all four against current code.

**#3 — `mobileChatInputRef` dead code → fixed** (`c523cd16`). Confirmed: declared, attached to the mobile FAB `<ChatInput>`, `.current` never read. The desktop `chatInputRef` *is* used (`getModelOptions()` for the edit/regen/retry rows), but the mobile FAB has none of those, so the ref was copy-paste residue. Removed the `useRef` + the `ref={}` prop (and the now-unused `ChatInputHandle` import).

**#4 — swallowed selection-commit error → fixed** (`c523cd16`). Agreed the bare `catch { /* ignore */ }` around the non-trivial `commitRegionSelection` (coord transforms + downsample) was the lone silent catch in the file. Now `console.debug`s the error with a comment scoping the expected case (chart disposed mid-gesture) vs. a real bug — matching the file's existing `console.debug`/`console.warn` catch convention, without re-introducing verbose logging.

**#1 — injection log lines → won't-fix (intentional).** Correct that the pre-existing `[PTC_CHAT] Directive context injected inline` / `Widget context injected inline` lines were dropped — but that was a deliberate call when `inject_inline_reminders` was consolidated (`35647bfa`), not an oversight. A consolidated log *inside* `inject_inline_reminders` is exactly the line that was removed on request. Leaving it out.

**#2 — `historyEventHandlers` double-read → won't-fix (deliberate).** The block (`hooks/utils/historyEventHandlers.ts:150-152`) mirrors the widget-restore block 8 lines above verbatim: the local is typed `unknown[]`, and the re-read of `event.metadata!.chart_selections` is the loosely-typed value that assigns to the message field without a cast. Using the local would either need a cast or diverge from the adjacent widget convention, for zero runtime difference.

Chen-zexi:
Verified all five against current code.

**Blocking — mobile FAB fast-mode cards → not blocking, misdiagnosed.** The mobile FAB has no message transcript to render onto: `MobileFabChat` renders only its `{children}` (the `<ChatInput>`), and `useMarketChat()` is consumed in exactly one place (`MarketView.tsx:192`), destructuring only `{ isLoading, handleSendMessage }` — `.messages` has zero consumers repo-wide. So fast-mode on mobile is fire-and-forget; there's no user bubble for cards to appear on. Forwarding `selectionSnapshots` into `handleFastModeSend` would store them on a `MarketChatMessage` that is never rendered (dead code). The agent side already works — the selection contexts ride `imageContext` (`457-460`, `487`). Your option 2 (switch the FAB to `useChatMessages`) means building a full mobile fast-mode transcript UI — a separate scope, not part of this feature.

**NB — `parse` raises → 500 → won't-fix.** The request field is typed `Optional[List[AdditionalContext]]` (`chat.py:236`), the discriminated union — malformed input is rejected at the boundary as **422, not a 500**, and valid items arrive as model instances that hit the `isinstance` branch, not `model_validate(dict)`. And `parse_widget_contexts` calls `WidgetContext.model_validate` unguarded too (`widget_context.py:56`), so this already matches the sibling parser you cited for consistency.

**NB — `price_level` equality → fixed** (`2115bc0b`). Correct gap. The geometry validator now collapses a stray `price_high` onto `price_low` for `price_level`, so the rendered level and the replayed metadata range can't disagree. Added a collapse test (and kept a region-ordering test) in `test_additional_context.py`.

**NB — `clearAll()` in PTC catch → won't-fix.** The dominant non-send path is the no-workspace early `return` (`499`), which *intentionally* keeps the selection so the user can create a workspace and retry. The `catch` only fires if `navigate()` throws — effectively never in react-router v6 — and on a genuinely failed send you want the chips preserved for retry, not wiped.

**NB — injection logging → won't-fix (intentional).** This is the third review to flag it; the removal was a deliberate call when `inject_inline_reminders` was consolidated, and it stays logging-free.

Chen-zexi:
Verified against current HEAD. Two of these are re-raises of items already dispositioned; the three new nits are fixed.

**#1 Blocking (mobile FAB cards) — already dispositioned (2×), not blocking.** Same misdiagnosis covered in issuecomment-4782907544 and issuecomment-4783523213: the mobile FAB has no transcript surface — `MobileFabChat` renders only its `{children}` (the `<ChatInput>`), and `useMarketChat().messages` has zero consumers repo-wide. Forwarding `selectionSnapshots` would store them on a never-rendered message. The agent already receives the selection via `imageContext`.

**#2 Logging — already dispositioned (3×), won't-fix.** Deliberate removal when `inject_inline_reminders` was consolidated; staying logging-free.

**#3 Cap divergence → fixed** (`d7bda68a`). Added a cross-ref comment on the client `MAX_SELECTION_BARS = 300` noting it must stay ≤ the server `_MAX_SELECTION_BARS`. (Harmless today since client < server, but the guard rail is cheap.)

**#5 dep churn → fixed** (`d7bda68a`). Keyed the outside-click effect on the primitive `activeKey` instead of the memo'd `active` object — matching the seed effect already in this file. `onDismiss` is a stable `useCallback([])`, so this fully removes the spurious re-subscriptions.

**#4 `_fmt` precision → fixed, but not as suggested** (`ed412a5b`). The `:.2f` suggestion would lose precision for sub-dollar / many-decimal instruments and shift any annotation the agent draws back. Used `%g` instead via a dedicated `_fmt_price` — magnitude-aware, so `195.10000000000002` renders as `195.1` while sub-dollar prices keep their precision. Applied to prices only (not volume — `%g` would scientific-notate large volumes). Added a noise-stripping test.

Chen-zexi:
**#1 Injection logging → won't-fix.** Same finding raised across several reviews; the removal was a deliberate call when `inject_inline_reminders` was consolidated, and it stays logging-free.

**#2 `console.debug` invisible in prod → fixed** (`2fa5a68b`). Fair point — `console.debug` is below the browser's default threshold, so the "surface it instead of swallowing" comment over-promised. Switched to `console.warn` with an honest comment. Used `warn` rather than `error`: a real `commitRegionSelection` bug would throw on every commit and never reach prod, so a throw here in prod is almost always the benign disposed-chart race — `warn` keeps it visible without being alert-grade.

**#3 unused `isConfirmedFor` import → stale.** `MarketChart.tsx` doesn't import `isConfirmedFor` — line 40 is `import { chartSelectionStore, useChartSelections } from '../stores/chartSelectionStore';`. The quoted import line isn't in this file; `isConfirmedFor` is imported and used in `SelectionCommentOverlay.tsx`. Nothing to remove.

Chen-zexi:
No changes this round — all three are won't-fix.

**#1 Injection logging → won't-fix.** Deliberate removal when `inject_inline_reminders` was consolidated; staying logging-free.

**#2 Builders run on skip paths → won't-fix.** Accurate that the `build_*_reminder()` calls execute before the helper sees the `None` target, but there's no real work to avoid: on HITL-resume and checkpoint-replay (regenerate/retry) the turn carries no new `additional_context`, so all three parsers return `[]` and each builder returns `None` immediately. And on checkpoint-replay specifically, `input_state` is `None` — there's no message list to inject into, and the original turn's reminders are already baked into the checkpoint being replayed. Guarding the hot path to skip three no-op calls would just split the skip logic across two places for no practical gain, so the single unconditional call stays.

**#3 Zero-height region → won't-fix.** A flat region (`price_low == price_high`) is a valid selection — a price level held across a time window — and the time bounds are exactly why it's a `region` rather than a `price_level`. Rejecting it would discard a deliberate drag; the degenerate draw-back hint (`point1.price == point2.price`) is an obvious flat rectangle, not a defect.

PR #283, opened 2026-06-25
feat(checkpointer): opt LangGraph messages channel into DeltaChannel

Chen-zexi:
## Summary

Opts the LangGraph `messages` channel into **`DeltaChannel`** (beta) so long
research threads store O(1)-per-step delta writes with a full snapshot every 50
updates, instead of re-serializing the entire message list every superstep
(O(N²) per thread over its life). Wired into the PTC, Flash, and subagent agent
factories via `state_schema=DeltaAgentState`.

This adopts **deepagents 0.6.11's clean DeltaChannel pattern verbatim**: a
**non-minting** messages reducer plus langgraph's upstream `ensure_message_ids`
(>=1.2.2), which stamps a stable id onto every id-less `messages` write *before*
it is persisted. No app-side id-stamping is needed — the reducer never re-mints on
replay, so reconstruction is deterministic by construction.

- **Performance** — `DeltaAgentState` + a vendored copy of deepagents 0.6.11's
  batch reducer (`state.py`); `state_schema` wired into all four `create_agent`
  call sites.
- **Determinism** — id stamping is langgraph's `ensure_message_ids`; the vendored
  reducer is non-minting (an id-less message is appended as-is, never re-rolled),
  so the hard-stop checkpoint flush / `/offload` write-back is a no-op and
  `RemoveMessage`-by-id stays stable across reconstruction.
- **Ops** — `scrub_nul_checkpoint.py` made `_DeltaSnapshot`-aware (preserves the
  NamedTuple on reserialize so the NUL-scrub maintenance script round-trips delta
  heads).
- **Guardrail** — `chat.py` role validator rejects roles unsafe under delta
  replay: those `convert_to_messages` can't rebuild (`tool`/`function`) AND those
  langgraph's `ensure_message_ids` doesn't stamp (`developer`) — an unstamped role
  persists id-less and would duplicate on the hard-stop flush.

**No data migration ships** — legacy `add_messages` threads reconstruct faithfully
under `DeltaChannel` with no backfill; see *Legacy-thread migration* below for the
investigation behind that.

## Overview

| Category               | Files | +LOC | −LOC |
|------------------------|------:|-----:|-----:|
| Modified (prod)        |     8 |   77 |    7 |
| New (prod)             |     1 |  116 |    0 |
| **Net (excl. tests + lock)** | **9** | **193** | **7** |
| Lock (uv.lock)         |     1 |   65 |   71 |
| Tests                  |     6 | 1113 |    1 |
| **Total**              | **16** | **1371** | **79** |

**By module**

- **Agent core** (`src/ptc_agent/`, net +123)
  - `agent/state.py` *(new, +116)* — `DeltaAgentState` + vendored **non-minting**
    `messages_delta_reducer` (dedup-by-id, `RemoveMessage` tombstones,
    `REMOVE_ALL_MESSAGES` reset, dict/str/tuple coercion, `None`-state replay
    guard); `MESSAGES_SNAPSHOT_FREQUENCY = 50`. Structurally identical to
    deepagents' `DeepAgentState`.
  - `agent/agent.py`, `agent/flash/agent.py`, `middleware/background_subagent/subagent.py`
    *(+2/+2/+3)* — pass `state_schema=DeltaAgentState` to the four `create_agent` calls.
- **Server** (`src/server/`, net +46/−2)
  - `utils/checkpointer.py` *(+6/−0)* — `get_checkpointer` returns plain
    `MemorySaver`/`AsyncPostgresSaver`; a comment records that langgraph's
    `ensure_message_ids` does the stamping, so no id-stamping shim is needed.
  - `handlers/workflow_handler.py` *(+0/−1)* — drops the vestigial
    `message_count` field from the `/status` response (see *Status field removed*).
  - `models/chat.py` *(+40/−1)* — `ChatMessage.role` validator: rejects roles
    unsafe under delta replay — `tool`/`function` (can't rebuild) and `developer`
    (converts but isn't stamped by langgraph, so it'd dup on the hard-stop flush).
- **Ops scripts** (`scripts/ops/`, net +9/−1)
  - `scrub_nul_checkpoint.py` *(+9/−1)* — preserve `_DeltaSnapshot` NamedTuple on reserialize.
- **Dependencies** — `pyproject.toml` *(+15/−4)* deepagents `>=0.6.11`, langgraph
  `>=1.2.2,<2`; `uv.lock` regen (langchain-core / langgraph cascade, plus a
  transitive `websockets` 16.0→15.0.1 downgrade pulled by langgraph-sdk/langsmith
  — within our `>=12.0` floor; `shared_ws_manager` uses a stable subset, so no
  code impact).
- **Tests** (`tests/unit/`, +1113) — see Test Coverage.

**Status field removed.** `GET /threads/{id}/status` previously returned a
`message_count` (`len(messages)` since the repo's initial-release squash). Verified
across all four services + frontend + integration gateway + CLI + docs: **no
consumer reads it** — the frontend's status type doesn't even declare it, and the
handler returns a plain untyped `dict` (it was never a typed contract). Under
DeltaChannel a non-snapshot step would need a ≤4-query delta-history walk to
reconstruct that count on ~98% of status polls, purely to populate a dead field.
Rather than carry that cost (the reviewers' P2), the field and its reconstruction
helper are removed outright — net effect on `workflow_handler.py` is a single
deleted line.

**What this introduces:** O(1)-per-step checkpoint storage for the messages
channel on long threads, made deterministic the same way deepagents does it
(non-minting reducer + langgraph's upstream id-stamp), plus a delta-aware NUL
scrub. Pre-existing threads need no migration — langgraph reconstructs their
legacy heads directly.

## Diff Size
- Total: 16 files changed, 1371 insertions(+), 79 deletions(−)
- Net (excl. tests + lock): 9 files changed, 193 insertions(+), 7 deletions(−)

## Test Coverage
Core behavior fully covered by unit tests; the legacy-thread reconstruction path
was additionally validated read-only against real production-shape checkpointer
data (see *Legacy-thread migration*).

Tests (6 files, all green — 4664 passed / 1 xpassed suite-wide; the lone
xfail/xpass is `test_fan_out_reconstruction_preserves_live_order`, a ~50/50
in-memory fan-out reordering documented as an accepted residual — both states green):
- `test_delta_channel_determinism.py` — **P1 regression guard** on a *plain*
  saver: id-less writes (BaseMessage + dict chat path) reconstruct to stable ids
  via `ensure_message_ids`; hard-stop flush is a no-op; `RemoveMessage`-by-id
  still removes; the async (`aput_writes`) path is covered.
- `test_messages_delta_reducer.py` — reducer parity vs deepagents 0.6.11's
  `_messages_delta_reducer` (drift guard) + semantic cases incl. the non-minting
  contract (id-less input stays id=None).
- `test_delta_channel_ordering.py` — same-superstep parallel ToolMessages: a
  reliable integrity guard (`test_fan_out_reconstruction_preserves_all_messages`,
  no message lost/duplicated) split out from the order comparison, which stays
  `xfail` as the accepted in-memory P2 residual.
- `test_delta_channel_resolution.py` — compiled graph's `messages` channel is
  `DeltaChannel` with `state_schema`, `BinaryOperatorAggregate` without.
- `test_scrub_nul_checkpoint.py` — `_DeltaSnapshot` survives NUL-scrub roundtrip.
- `test_chat_models.py` — role validator: `developer` rejected (converts but
  unstamped → would dup on flush) + `test_every_valid_role_is_stamped_by_langgraph`
  ties the accepted set to langgraph's `_MESSAGE_ROLES` so a re-add or langgraph
  bump fails loudly.

## Conformance with deepagents

Adopts deepagents 0.6.11's DeltaChannel pattern as-is — no divergence, no extra
machinery.

**Same as deepagents (parity-tested):**
- **Reducer** — `messages_delta_reducer` (`state.py`) is a vendored copy of
  deepagents 0.6.11's `_messages_delta_reducer`: dedup-by-id, `RemoveMessage`
  tombstones, `REMOVE_ALL_MESSAGES` reset, dict/str/tuple coercion, and
  **non-minting** (id-less messages appended as-is; ids come from langgraph's
  `ensure_message_ids`). `test_vendored_reducer_matches_deepagents` fails CI if our
  copy drifts from the installed symbol — including byte-for-byte details like the
  `state_msgs` coercion line, which we keep identical to upstream.
- **Schema** — `DeltaAgentState` equals deepagents' `DeepAgentState`: same
  `AgentState` base, same `Required[Annotated[list[AnyMessage], DeltaChannel(reducer,
  snapshot_frequency=50)]]` field.

**Only difference — vendored, not imported.** We keep our own reducer copy rather
than importing the private `deepagents._messages_reducer` symbol, so on-disk
reconstruction stays frozen to our release (a beta API can move without a semver
bump); the parity test bridges the two. We also wire delta via `create_agent` +
`state_schema=` rather than `create_deep_agent`, keeping our own middleware stack
and backend filesystem — same delta wiring on `messages`.

**No id-stamping shim.** An earlier revision of this branch carried a checkpointer
shim to stamp id-less `Overwrite`-wrapped writes (the one path `ensure_message_ids`
misses). It's gone: deepagents 0.6.11 changed the only middleware that emitted that
shape — `PatchToolCallsMiddleware`, repairing a dangling tool call after a mid-tool
cancel — to write a plain list that `ensure_message_ids` covers, and langalpha
emits no id-less `Overwrite` writes of its own. With a non-minting reducer, even a
hypothetical miss reconstructs deterministically (id stays None) rather than
duplicating, so the shim and its tests earned removal.

## Legacy-thread migration: investigated, not needed

An earlier revision of this branch shipped a one-shot backfill script on the
assumption that threads checkpointed under the old `add_messages` reducer would
reconstruct incorrectly once `messages` became a `DeltaChannel` — losing message
ids, reordering, or duplicating on the first post-deploy read. **That assumption
was wrong; the backfill (and its tests) have been removed.** Recording the analysis
so the decision is auditable and reviewers know why no migration ships:

- **The wrong direction.** The "threads diverge" conclusion came from exercising the
  low-level delta replay method (`aget_delta_channel_history(..., seed=False)`)
  *directly*. That method does walk raw ancestor `checkpoint_writes` and, for writes
  persisted before langgraph's `ensure_message_ids` stamped them, will drop ids /
  skip dedup / reorder. But that is **not the path production uses** to load a
  thread.

- **The correction.** `DeltaChannel.from_checkpoint` has an explicit migration
  branch: when a stored `messages` blob is a plain list (a legacy
  `BinaryOperatorAggregate` value) rather than a `_DeltaSnapshot` or sentinel, the
  channel **uses that list directly** — the reducer is never invoked and no ancestor
  writes are replayed. A legacy `add_messages` thread materializes the full list in
  *every* checkpoint, including its head, so loading it under `DeltaChannel` reads
  that head verbatim. Confirmed read-only via the production `graph.aget_state` path
  against real legacy threads — including large, deeply-checkpointed,
  multi-hundred-message histories — and corroborated by runtime instrumentation
  showing every legacy head took the plain-value branch with **zero** threads on the
  replay path and **zero** id-less messages in any real head.

- **Why the backfill was not just unnecessary but harmful.** It re-snapshotted each
  legacy head via `aupdate_state`. In langgraph, `update_state` / `bulk_update_state`
  call `create_checkpoint(...)` **without** `channels_to_snapshot` (only the runtime
  superstep path supplies it), so a `DeltaChannel` value is *omitted* from the new
  head — a sentinel, **not** a `_DeltaSnapshot`. That **forces** the very
  ancestor-replay path the migration was meant to avoid, turning a faithful legacy
  head into a divergent reconstruction (id loss + duplication). Reproduced on a real
  `AsyncPostgresSaver`.

- **Subagents.** Background-subagent lineages (`task:<id>` checkpoint namespaces) are
  the same: legacy heads reconstruct directly and faithfully. They were never
  addressable by a standalone re-snapshot graph anyway (a non-root `checkpoint_ns`
  is reset to root on a standalone `aupdate_state`), so nothing here needed — or
  could — backfill them.

**Net:** the DeltaChannel opt-in is self-migrating by langgraph's design. No data
migration and no deploy-ordering constraint for existing threads.

## Rollout notes
1. **No data migration / no deploy ordering** (see *Legacy-thread migration*) —
   existing threads reconstruct faithfully on first load.
2. **One-way data format.** Once a thread takes a step under `DeltaChannel`, its head
   becomes a sentinel or `_DeltaSnapshot` that reverting to `add_messages` (or
   downgrading langgraph below 1.2) cannot read — keep the `langgraph` /
   `langgraph-checkpoint*` floors pinned at `>=1.2`. Validate on staging first.
3. **deepagents floor raised to >=0.6.11.** The non-minting reducer parity and the
   no-shim assumption both require it; the bump pulls a minor langchain-core
   (1.4.0→1.4.8) / langgraph (1.2.4→1.2.6) cascade. Full unit suite green on it.

## Test plan
- [x] `uv run pytest tests/unit/` — 4664 passed, 1 xpassed
- [x] Verified read-only that real legacy `add_messages` threads reconstruct
      faithfully under `DeltaChannel` via the production `graph.aget_state` path
      (same length / ids / order / contents; zero threads on the replay path) — so
      no migration is required
- [x] Live E2E (running backend, real Postgres): hard-stop checkpoint flush →
      reconstruct ×2 → no dup, ids stable; resume turn appends cleanly;
      `role:"developer"` → 422
- [x] `uv run ruff check` on changed files — clean
- [ ] Staging: resume a pre-change thread, hard-stop flush, `/offload`,
      multi-tool-call turn; verify `checkpoint_blobs` are sentinels on non-snapshot
      steps, full `_DeltaSnapshot` every ~50 updates.

## Summary by CodeRabbit

* **New Features**
  * Added delta-based message checkpointing support for agent state, improving consistency when messages are saved and restored.
  * Added stricter validation for chat message roles, rejecting unsupported or misspelled roles earlier.

* **Bug Fixes**
  * Preserved message types and ordering during checkpoint cleanup and reconstruction.
  * Improved handling of message IDs so removals and replay behave more reliably.
  * Removed an outdated checkpoint message count from workflow status details.

* **Tests**
  * Added broad regression coverage for message reconstruction, ordering, cleanup, and role validation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,9 +18,15 @@ dependencies = [
     "langchain-anthropic>=1.0.0",
     "langchain_google_genai>=2.0.0",
     "langchain-deepseek>=0.1.4",
-    # LangGraph
-    "langgraph>=0.3.5",
-    "langgraph-checkpoint-postgres>=2.0.0",
+    # LangGraph — floor pinned at the beta DeltaChannel on-disk format (messages
+    # checkpoints are _DeltaSnapshot/sentinel blobs). langgraph<1.2 cannot read
+    # them; downgrading after delta blobs exist strands threads on resume.
+    # >=1.2.2 brings upstream id-stamping for DeltaChannel writes (ensure_message_ids
+    # in pregel/_messages.py), which stamps id-less BaseMessage/dict/list writes
+    # before persistence — the mechanism that keeps reconstruction deterministic
+    # paired with the non-minting reducer in ptc_agent.agent.state.
+    "langgraph>=1.2.2,<2",
+    "langgraph-checkpoint-postgres>=3.1",
     # Other dependencies
     "python-dotenv>=1.0.1",
     # Pinned: we import from scrapling.engines._browsers._controllers /
@@ -64,7 +70,12 @@ dependencies = [
     "edgartools>=5.7.2",
     "langsmith-fetch>=0.3.1",
     "structlog",
-    "deepagents>=0.5.3",
+    # Floor at 0.6.11: that release made the messages delta reducer non-minting
+    # (our vendored copy in ptc_agent.agent.state matches it — see the parity test)
+    # and switched PatchToolCallsMiddleware off the id-less `Overwrite` write, so
+    # langgraph's ensure_message_ids covers every id-less messages write and no
+    # app-side id-stamping shim is needed.
+    "deepagents>=0.6.11",
     "daytona-sdk>=0.130.0",
     "boto3>=1.42.28",
     "tavily-python>=0.7.18",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,9 @@ dependencies = [
     # before persistence — the mechanism that keeps reconstruction deterministic
     # paired with the non-minting reducer in ptc_agent.agent.state.
     "langgraph>=1.2.2,<2",
-    "langgraph-checkpoint-postgres>=3.1",
+    # ceiling mirrors langgraph<2 — the postgres saver serializes the same delta
+    # blobs and moves in lockstep with core, so it can't cross a major on its own.
+    "langgraph-checkpoint-postgres>=3.1,<4",
     # Other dependencies
     "python-dotenv>=1.0.1",
     # Pinned: we import from scrapling.engines._browsers._controllers /
--- a/scripts/ops/scrub_nul_checkpoint.py
+++ b/scripts/ops/scrub_nul_checkpoint.py
@@ -67,7 +67,15 @@ def _strip_nul(value: Any) -> tuple[Any, int]:
         return out_list, n
     if isinstance(value, tuple):
         cleaned = [_strip_nul(item) for item in value]
-        return tuple(c for c, _ in cleaned), sum(n for _, n in cleaned)
+        fields = [c for c, _ in cleaned]
+        total = sum(n for _, n in cleaned)
+        # NamedTuple (e.g. langgraph's `_DeltaSnapshot`): reconstruct with the
+        # SAME type so the serializer still emits its dedicated ext code.
+        # Flattening to a plain tuple would make `DeltaChannel.from_checkpoint`
+        # stop recognizing the snapshot, corrupting the channel on resume.
+        if hasattr(value, "_fields"):
+            return type(value)(*fields), total
+        return tuple(fields), total
     # Pydantic/LangChain message objects: walk attributes that look textual.
     # `content` is the dominant carrier. Other str attributes get same treatment.
     if hasattr(value, "__dict__"):
--- a/src/ptc_agent/agent/agent.py
+++ b/src/ptc_agent/agent/agent.py
@@ -21,6 +21,7 @@ from ptc_agent.agent.backends import (
     StoreBackend,
 )
 from ptc_agent.agent.middleware import SubAgentMiddleware
+from ptc_agent.agent.state import DeltaAgentState
 from deepagents.middleware.patch_tool_calls import PatchToolCallsMiddleware
 from langchain_anthropic.middleware import AnthropicPromptCachingMiddleware
 
@@ -750,6 +751,7 @@ class PTCAgent:
             middleware=deepagent_middleware,
             checkpointer=checkpointer,
             store=store,
+            state_schema=DeltaAgentState,
         ).with_config({"recursion_limit": 2000})
 
         return BackgroundSubagentOrchestrator(
--- a/src/ptc_agent/agent/flash/agent.py
+++ b/src/ptc_agent/agent/flash/agent.py
@@ -25,6 +25,7 @@ from ptc_agent.agent.middleware import (
     ProvenanceMiddleware,
 )
 from ptc_agent.agent.middleware.runtime_context import RuntimeContextMiddleware
+from ptc_agent.agent.state import DeltaAgentState
 from ptc_agent.agent.prompts import format_current_time, get_loader
 from ptc_agent.config import AgentConfig
 
@@ -334,6 +335,7 @@ class FlashAgent:
             middleware=middleware,
             checkpointer=checkpointer,
             store=store,
+            state_schema=DeltaAgentState,
         )
         if response_format is not None:
             create_kwargs["response_format"] = response_format
--- a/src/ptc_agent/agent/middleware/background_subagent/subagent.py
+++ b/src/ptc_agent/agent/middleware/background_subagent/subagent.py
@@ -26,6 +26,7 @@ from ptc_agent.agent.middleware.background_subagent.middleware import (
 )
 from ptc_agent.agent.middleware.background_subagent.registry import BackgroundTaskRegistry
 from ptc_agent.agent.middleware._utils import append_to_system_message
+from ptc_agent.agent.state import DeltaAgentState
 from src.llms.content_utils import extract_reasoning_summary_index
 from src.llms.llm import narrow_prompt_cache_key
 from src.server.utils.content_normalizer import normalize_text_content
@@ -454,6 +455,7 @@ def _get_subagents(
             middleware=general_purpose_middleware,
             name="general-purpose",
             checkpointer=checkpointer,
+            state_schema=DeltaAgentState,
         )
         agents["general-purpose"] = general_purpose_subagent
         subagent_descriptions.append(
@@ -488,6 +490,7 @@ def _get_subagents(
             middleware=_middleware,
             name=agent_["name"],
             checkpointer=checkpointer,
+            state_schema=DeltaAgentState,
         )
     return agents, subagent_descriptions
 
--- a/src/ptc_agent/agent/state.py
+++ b/src/ptc_agent/agent/state.py
@@ -1,0 +1,116 @@
+"""DeltaChannel state schema for the messages key.
+
+Vendors deepagents' batch reducer as a public `messages_delta_reducer` and
+defines `DeltaAgentState` (an `AgentState` whose `messages` uses `DeltaChannel`
+for O(1)-per-step checkpoint storage instead of re-serializing the full list).
+`DeltaAgentState` is structurally identical to deepagents 0.6.11's
+`DeepAgentState` (`AgentState` + `DeltaChannel(reducer, snapshot_frequency=50)`);
+we vendor the reducer rather than import the private `deepagents._messages_reducer`
+symbol so on-disk reconstruction stays frozen to our release. The parity test
+`test_messages_delta_reducer.py::test_vendored_reducer_matches_deepagents` fails
+CI if our copy drifts.
+
+One-way data format: once a thread is checkpointed under `DeltaChannel` its head
+blob is a sentinel or `_DeltaSnapshot`, which reverting to `add_messages` (or
+downgrading langgraph below 1.2) cannot read — so keep the `langgraph`/
+`langgraph-checkpoint*` floors pinned at >=1.2 in pyproject.toml.
+"""
+
+from __future__ import annotations
+
+from typing import Annotated, Any, cast
+
+from langchain.agents import AgentState
+from langchain_core.messages import (
+    AnyMessage,
+    BaseMessage,
+    RemoveMessage,
+    convert_to_messages,
+)
+from langgraph.channels import DeltaChannel
+from langgraph.graph.message import REMOVE_ALL_MESSAGES
+from typing_extensions import Required
+
+# A full snapshot blob is written every N updates, bounding delta replay depth
+# (matches deepagents' tested default). Single source of truth for the
+# `DeltaChannel` snapshot frequency, consumed by `DeltaAgentState` below.
+MESSAGES_SNAPSHOT_FREQUENCY = 50
+
+
+def messages_delta_reducer(  # noqa: C901, PLR0912
+    state: list[AnyMessage], writes: list[list[AnyMessage]]
+) -> list[AnyMessage]:
+    """Batch reducer for `DeltaChannel` on the messages key.
+
+    Dedups by id, tombstones via `RemoveMessage`, resets on
+    `REMOVE_ALL_MESSAGES`, and coerces raw dict/str/tuple input via
+    `convert_to_messages`. IDs are NOT assigned here: langgraph's
+    `ensure_message_ids` (>=1.2.2) stamps id-less writes before they are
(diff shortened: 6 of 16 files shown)


Chen-zexi:
Pushed a follow-up. Disposition of all review findings:

**Fixed**
- **`message_count` polling cost (P2)** — removed the field outright rather than caching it. Verified no consumer reads it across all four services + frontend + integration gateway + CLI + docs: the frontend status type doesn't even declare it, and the handler returns an untyped `dict` (never a typed contract). It existed since the repo's initial-release squash as `len(messages)`; under DeltaChannel a non-snapshot step needed a ≤4-query delta-history walk to populate a dead field on ~98% of polls. `_delta_safe_message_count` and its test are gone; net effect on `workflow_handler.py` is one deleted line.
- **`backfill_delta.py` enumeration failure → `None` → exit 0** — now raises `DeltaBackfillError` so `main` exits non-zero, and a deploy gated on the backfill can't proceed on a swallowed DB error. Test added.
- **xfail masking in `test_delta_channel_ordering.py`** — split the no-loss/no-duplicate integrity assertion into its own non-xfail test; only the in-memory order comparison stays `xfail`. A real message-loss regression can no longer hide as XFAIL.

**Mooted**
- Hard-coded `50` vs `MESSAGES_SNAPSHOT_FREQUENCY` — that test file was removed with the field.

**Won't fix (with reason)**
- `state.py` long line — deepagents upstream is byte-identical there; wrapping would diverge our vendored copy from the parity source it's diffed against. Ruff is clean (E501 not enabled).
- backfill "fail closed when registry unavailable" — the registry is process-local and legitimately absent when the script runs standalone (its own `main` builds the checkpointer), so failing closed would skip *every* lineage. The checkpoint-id + `pending_writes` re-read fallback plus the documented "run drained" guidance bound the residual sub-ms window intentionally.
- parity helper checks only type/content/id-presence — the reducer never reconstructs messages (it dedups/appends existing objects), so `tool_call_id` etc. can't be stripped via this path; `test_vendored_reducer_matches_deepagents` is the behavioral drift guard.

**Left as-is (trivial docs, already enforced)** — "note new roles need opt-in" / "document the private `_MESSAGE_ROLES` coupling": `test_every_valid_role_is_stamped_by_langgraph` already fails CI on a bad re-add, and the test docstring already documents the coupling.

Suite: 4670 passed / 1 xpassed; ruff clean.

Chen-zexi:
Pushed. This round:

**Fixed**
- **Integrity test used `set()`**, which masks a *duplicated* reconstruction id — the exact corruption that guard exists to catch (`[seed, A, A]` passes against `{seed, A}`). Now `Counter(recon_ids) == Counter(live_ids)`: multiset equality catches both loss and duplication, while order stays the separate `xfail`.
- **`backfill_delta.py` silently swallowed the active-task-registry import failure** — added a `logger.debug` so a first-run "why didn't the active guard fire?" is traceable. Behavior unchanged; the standalone fallback is intentional.

**Won't fix (with reason)**
- **`scrub_nul_checkpoint.py` `type(value)(*fields)` defensive fallback** — the suggested `tuple(fields)` fallback would drop the NamedTuple type langgraph expects, a worse corruption than failing loud. Only `_DeltaSnapshot` (trivial `__new__`) is handled; the failure mode is hypothetical and the proposed fix is wrong for an ops script.
- **`state.py` noqa comment** — deepagents upstream carries the identical `# noqa: C901, PLR0912` on that line; it's part of the byte-identical vendored copy. An inline comment would diverge the very line it's meant to protect, and the vendoring intent is already documented in the module + function docstrings where a refactorer would look.

Suite: ruff clean; ordering + migration green (9 passed / 1 xfailed).

Chen-zexi:
Both items addressed.

**Fixed — stale comment.** Dropping `message_count` left `state.py`'s `MESSAGES_SNAPSHOT_FREQUENCY` comment claiming `workflow_handler` "MUST use this … keep in sync with both call sites," but that consumer is gone — only one remains. Rewrote it to name the single remaining consumer (`DeltaAgentState`). Comment-only; `workflow_handler.py` stays net `+0/−1`.

**No-impact, documented — the `websockets` 16.0→15.0.1 downgrade.** It's transitive, pulled by langgraph-sdk 0.4.2's new `websockets<16` cap (not our floor — we're `>=12.0`). The only backwards-incompatible change in 16.0 is dropping Python 3.9; we run 3.12, so it's moot. `shared_ws_manager` uses the asyncio client surface (`websockets.connect` + `websockets.exceptions.ConnectionClosed`), unchanged across 15/16 — 16.0's deltas are additive (per-fragment `max_size`, HTTP/1.0 proxy, `Server.close` opts) or fixes on paths we don't hit (threading-impl close race — we're asyncio; a `recv()` bytearray edge case). `backend-integration` (the live WS fan-out) is green on 15.0.1. Noted in the PR body's Dependencies line; it floats back to 16 with no code change once the sdk cap lifts.

Chen-zexi:
This review ran against a stale snapshot of the description (it fired on the force-push; the body was rewritten ~8 min before this comment landed). Re-grounding on the current state:

### 🚫 Blocking — resolved

`backfill_delta.py` / `test_delta_migration.py` aren't missing — they were **cut deliberately**, and the description now matches the commit (16 files / 1371 insertions, consistent with the API; zero references to `backfill_delta` or `--apply`). See the new **"Legacy-thread migration: investigated, not needed"** section. There is no command for an operator to run, because no migration ships.

Why the script is gone rather than deferred:

- **Legacy `add_messages` threads reconstruct faithfully under `DeltaChannel` with no migration.** `DeltaChannel.from_checkpoint` has a migration branch: when a stored `messages` blob is a plain list (the legacy `BinaryOperatorAggregate` value) rather than a `_DeltaSnapshot`/sentinel, the channel uses it directly — no reducer, no ancestor replay. Every checkpoint, including the head, materializes the full list as stored.
- **The backfill was not just unnecessary but harmful.** Its `aupdate_state` path calls `create_checkpoint(...)` without `channels_to_snapshot`, so it writes a **sentinel** head, not a snapshot. That forces id-less ancestor replay on read — turning a faithful thread into a divergent one (id loss + duplication). Shipping it would have *introduced* the corruption the P1 note feared.
- **The original "P1 deployment dependency" framing came from testing the low-level replay method** (`aget_delta_channel_history(seed=False)`) directly, which is not the production `graph.aget_state` path for plain-list-head threads. Re-grounding the test on the real read path reversed the conclusion.

Subagents behave identically — their legacy heads reconstruct faithfully too, so no migration is needed there either.

### 💡 Non-blocking — agreed

Good catch. Mirroring the `langgraph>=1.2.2,<2` ceiling, I'm pinning `langgraph-checkpoint-postgres>=3.1,<4` so the saver can't cross a major independently of core. Lands in the next push.

PR #286, opened 2026-06-26
fix(docker): add docker sandbox rootless local setup

alexferrari88:
## Summary

This isolates the local Docker sandbox fixes needed for rootless Docker environments.

- expose `DOCKER_HOST_SOCKET` in `.env.example` so rootless Docker users can mount `/run/user/<uid>/docker.sock` into the backend
- mount `Dockerfile.sandbox` into the backend container so the Docker provider can auto-build `langalpha-sandbox:latest`
- build missing sandbox images through `aiodocker` with a tar stream (`fileobj` + `path_dockerfile`) instead of Docker SDK-style `path` / `dockerfile` kwargs
- add unit coverage for the `aiodocker` image build call shape

## Why

When `SANDBOX_PROVIDER=docker` runs inside the compose backend, sandbox creation can fail before a container is created if the backend cannot see the correct Docker socket or cannot find `Dockerfile.sandbox` at `/app/Dockerfile.sandbox`. With rootless Docker, the host socket commonly lives under `/run/user/<uid>/docker.sock`, so the compose mount needs to be configurable.

The image auto-build path also needs to match `aiodocker 0.26`, which expects a tar stream and `path_dockerfile`; passing Docker SDK kwargs fails during sandbox image creation.

## Validation

- `uv run --extra dev ruff check src/ptc_agent/core/sandbox/providers/docker.py tests/unit/core/sandbox/test_docker_provider.py`
- `uv run pytest tests/unit/core/sandbox/test_docker_provider.py -q` (`111 passed`)
- `docker compose config -q`

## Summary by CodeRabbit

* **New Features**
  * Added a `DOCKER_HOST_SOCKET` configuration option (including rootless Docker guidance) and updated the development container to use it for the Docker host socket.
* **Bug Fixes**
  * Improved Docker image auto-build behavior by correctly building from an in-memory Dockerfile archive, improving reliability when the target image is missing.
* **Tests**
  * Expanded unit test coverage for the Docker image build fallback path and refined related assertions for container startup and host resolution.

### DIFF
--- a/src/ptc_agent/core/sandbox/providers/docker.py
+++ b/src/ptc_agent/core/sandbox/providers/docker.py
@@ -877,7 +877,7 @@ class DockerProvider(SandboxProvider):
                 f"Build the image manually or place {dockerfile_name} in the repo root."
             )
 
-        build_context = os.path.dirname(dockerfile_path)
+        dockerfile_basename = os.path.basename(dockerfile_path)
         image_tag = self._config.image
 
         logger.info(
@@ -886,18 +886,17 @@ class DockerProvider(SandboxProvider):
             tag=image_tag,
         )
 
-        # Use aiodocker to build. aiodocker 0.26 does not accept the Docker SDK's
-        # ``path=``/``dockerfile=`` kwargs; it expects a tar stream plus
-        # ``path_dockerfile``.
+        # Use aiodocker to build. It expects a tar stream via ``fileobj`` and
+        # the Dockerfile path via ``path_dockerfile``.
         try:
             build_tar = io.BytesIO()
             with tarfile.open(fileobj=build_tar, mode="w") as tar:
-                tar.add(build_context, arcname=".")
+                tar.add(dockerfile_path, arcname=dockerfile_basename)
             build_tar.seek(0)
 
             async for log_line in client.images.build(
                 fileobj=build_tar,
-                path_dockerfile=os.path.basename(dockerfile_path),
+                path_dockerfile=dockerfile_basename,
                 tag=image_tag,
                 rm=True,
                 stream=True,
--- a/src/ptc_agent/core/sandbox/providers/docker.py
+++ b/src/ptc_agent/core/sandbox/providers/docker.py
@@ -886,14 +886,21 @@ class DockerProvider(SandboxProvider):
             tag=image_tag,
         )
 
-        # Use aiodocker to build
+        # Use aiodocker to build. aiodocker 0.26 does not accept the Docker SDK's
+        # ``path=``/``dockerfile=`` kwargs; it expects a tar stream plus
+        # ``path_dockerfile``.
         try:
-            # aiodocker build expects a tar context or path
+            build_tar = io.BytesIO()
+            with tarfile.open(fileobj=build_tar, mode="w") as tar:
+                tar.add(build_context, arcname=".")
+            build_tar.seek(0)
+
             async for log_line in client.images.build(
-                path=build_context,
-                dockerfile=os.path.basename(dockerfile_path),
+                fileobj=build_tar,
+                path_dockerfile=os.path.basename(dockerfile_path),
                 tag=image_tag,
                 rm=True,
+                stream=True,
             ):
                 if isinstance(log_line, dict) and "stream" in log_line:
                     line = log_line["stream"].strip()
--- a/src/ptc_agent/core/sandbox/providers/docker.py
+++ b/src/ptc_agent/core/sandbox/providers/docker.py
@@ -900,6 +900,7 @@ class DockerProvider(SandboxProvider):
                 tag=image_tag,
                 rm=True,
                 stream=True,
+                encoding="identity",
             ):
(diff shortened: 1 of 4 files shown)


src/ptc_agent/core/sandbox/providers/docker.py
```diff
@@ -886,14 +886,21 @@ async def _ensure_image(self, client: Any) -> None:
...
+            build_tar.seek(0)
+
             async for log_line in client.images.build(
-                path=build_context,
-                dockerfile=os.path.basename(dockerfile_path),
+                fileobj=build_tar,
+                path_dockerfile=os.path.basename(dockerfile_path),
                 tag=image_tag,
                 rm=True,
+                stream=True,
```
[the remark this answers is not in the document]
alexferrari88: Fixed in e15f1849. I removed the version-specific `aiodocker 0.26` wording and kept the comment focused on the actual supported call shape: `fileobj` tar stream plus `path_dockerfile`.

alexferrari88:
Follow-up pushed in e15f1849:

- Removed the version-specific `aiodocker 0.26` comment so it matches the project dependency range.
- Reduced the sandbox image build context to a tar containing only `Dockerfile.sandbox`; the current Dockerfile has no `COPY`/`ADD` inputs that require the full app directory.
- Extended the unit test to assert the tar stream excludes unrelated files.

Validation:
- `uv run --extra dev ruff check src/ptc_agent/core/sandbox/providers/docker.py tests/unit/core/sandbox/test_docker_provider.py`
- `uv run pytest tests/unit/core/sandbox/test_docker_provider.py -q` (`111 passed`)

src/ptc_agent/core/sandbox/providers/docker.py
```diff
@@ -886,14 +886,20 @@ async def _ensure_image(self, client: Any) -> None:
...
-            # aiodocker build expects a tar context or path
+            build_tar = io.BytesIO()
+            with tarfile.open(fileobj=build_tar, mode="w") as tar:
+                tar.add(dockerfile_path, arcname=dockerfile_basename)
+            build_tar.seek(0)
+
             async for log_line in client.images.build(
-                path=build_context,
-                dockerfile=os.path.basename(dockerfile_path),
+                fileobj=build_tar,
```
Chen-zexi, line 898: aiodocker 0.26's `build()` raises `ValueError("You need to specify an encoding")` whenever `fileobj` is set without `encoding`, and that guard runs synchronously — before any call to the daemon. So as written this `build()` raises before the image is ever built. Please pass `encoding="identity"` here (the tar is uncompressed — `mode="w"`):

```python
            async for log_line in client.images.build(
                fileobj=build_tar,
                path_dockerfile=dockerfile_basename,
                tag=image_tag,
                rm=True,
                stream=True,
                encoding="identity",
            ):
```

Confirmed end-to-end against a real daemon: the current kwargs fail with that `ValueError`; adding `encoding="identity"` builds successfully.

Chen-zexi (changes requested):
Thanks for the PR.

Heads-up on the validation: the `111 passed` run doesn't actually exercise the auto-build — the test mocks `client.images.build`, so aiodocker's real argument validation never runs. That's how the issue below stays green.

Two asks before merge:

1. **(blocking)** The `images.build(...)` call is missing the required `encoding` argument — see the inline comment. aiodocker 0.26 raises `ValueError("You need to specify an encoding")` synchronously whenever `fileobj` is set without it, so the auto-build path this PR adds fails before the daemon is even contacted (it surfaces as `RuntimeError: Failed to build Docker image 'langalpha-sandbox:latest': You need to specify an encoding`). I reproduced it end-to-end against a real Docker daemon: the current kwargs raise that `ValueError`; adding `encoding="identity"` builds successfully. Pre-PR `main` was broken here too (`path=`/`dockerfile=` aren't valid aiodocker 0.26 kwargs), so this is the last step to actually make auto-build work.
2. Please extend `test_ensure_image_builds_with_aiodocker_tar_stream` to assert `build_kwargs["encoding"] == "identity"` (or exercise one real build). With `images.build` mocked the test passes even when the real call can't run — which is exactly what happened here.

src/ptc_agent/core/sandbox/providers/docker.py
alexferrari88: Fixed in alexferrari88/LangAlpha@27b44532.

- Added `encoding="identity"` to the `client.images.build(...)` call because the tar stream is uncompressed.
- Extended `test_ensure_image_builds_with_aiodocker_tar_stream` to assert `build_kwargs["encoding"] == "identity"`.

Validation:
- `uv run --extra dev ruff check src/ptc_agent/core/sandbox/providers/docker.py tests/unit/core/sandbox/test_docker_provider.py`
- `uv run pytest tests/unit/core/sandbox/test_docker_provider.py -q` (`111 passed`)

Chen-zexi (approved):

Chen-zexi:
Thanks for the PR, @alexferrari88 — merged in 9ffccdd. Nice work.

PR #288, opened 2026-06-26
feat(provenance): capture bash-run MCP calls + store full per-access result bodies

Chen-zexi:
## Overview

Two changes to the provenance pipeline, relative to `main`:

1. **New capture surface — MCP calls from bash-run scripts.** Provenance now
   traces MCP tool calls made by a Python script run via `Bash` (foreground) and
   by a backgrounded script surfaced through `BashOutput`; previously only direct
   `execute_code` calls were traced, so MCP access from `python analysis.py`-style
   runs was invisible to the Sources panel and to a verifier. Closing this needed
   sandbox-side wiring: the bash command now exports `MCP_TRACE_FILE` and prepends
   the wrapper path to `PYTHONPATH`, so a plain `python script.py` resolves the
   generated MCP client and writes the same trace `execute_code` does.
2. **Full result bodies.** Provenance kept only a redacted snippet +
   `result_sha256` per access, so a verifier couldn't read the bytes the agent
   reasoned over. This adds a global, content-addressed **body store**: the raw
   per-access result captured at the access site, deduplicated by content hash,
   redacted before storage, spilled to object storage above 64 KiB, with read
   endpoints that hand back the exact, hash-consistent body.

| Category            | Files | +LOC | −LOC |
|---------------------|------:|-----:|-----:|
| Modified (prod)     |    11 |  700 |   43 |
| New (prod)          |     3 |  537 |    0 |
| Tests (excluded)    |    14 | 2,960 |    6 |
| **Net (excl. tests)** | **14** | **1,237** | **43** |

### migrations  (net +64)
- `015_add_provenance_result_bodies.py` (+64) — new `provenance_result_bodies`
  table (`result_sha256` PK, `body_inline`, `object_key`, true `byte_len`,
  `content_type`, `created_at`/`last_seen_at`) + `idx_provenance_records_sha`
  serving the GC mark-sweep and verifier joins. No FK — bodies are shared and
  GC'd independently of records.

### agent — capture & trace  (net +387)
- `core/sandbox/ptc_sandbox.py` (+150/−5) — closes the bash provenance bypass: a
  shared `_build_trace_env_command` injects `MCP_TRACE_FILE` + a `PYTHONPATH`
  prepend into both foreground and background bash execution so a `python
  script.py` resolves the MCP wrappers and writes a trace; the foreground path
  collects that trace (with recovery on exception and `asyncio.shield` cleanup on
  cancel), and backgrounded traces are tracked per command and harvested on the
  `BashOutput` that observes completion (dropped + logged if the session is
  evicted unobserved). `_collect_mcp_trace` enforces a **16 MiB host-side read
  cap** — the hard memory bound against agent code appending to the trace file
  directly (the codegen budget below is only cooperative). `_bg_trace_paths` is
  cleared on `stop()`/`cleanup()` and dropped on `stop_background_command`.
- `agent/middleware/provenance/middleware.py` (+182/−9) — registers `Bash` and
  `BashOutput` as provenance sources alongside `ExecuteCode` (shared
  `_extract_execute_code`, `mcp_trace` strip, error-exemption via
  `_MCP_TRACE_TOOLS`); each `_extract_*` stamps `result_body` via `_canonical_body`
  from the **same** value it hashes (per-item for fan-out; `execute_code` pulls
  each mcp_tool call's own trace body via `_clamp_body_bytes`, never the aggregate
  stdout); `_verify_source_sha` trust-gates agent-authored `mcp_tool` shas —
  nulling both sha and body on mismatch (so the emitted SSE event also carries a
  null sha) and re-sizing `result_size` from the verified bytes on match;
  `_body_item`/`_redact_body`/`_flush_bodies` batch one redacted store-write **per
  tool call** (collapsing that call's fan-out). The SSE event shape is unchanged —
  no body bytes ride it.
- `core/tool_generator.py` (+36/−1) — generated MCP client emits a byte-capped
  `result_body` per call plus an aggregate per-execution budget (~4 MiB,
  cooperative) after which it stops emitting bodies (sha/snippet/size still
  recorded). Codegen version bumped to force the new client onto reused sandboxes.
- `agent/tools/bash.py` (+20/−8), `agent/tools/bash_output.py` (+19/−8) — return
  `(content, {"mcp_trace": [...]})` so the trace reaches the middleware; the
  middleware strips it before the artifact leaves the host (never enters the LLM
  context).
- `agent/provenance/types.py` (+11) — `result_body` field on `ProvenanceSource`
  + canonical `RESULT_BODY_MAX_BYTES = 64 KiB`.

### server — store, serve, GC  (net +691)
- `database/provenance_bodies.py` (+378, new) — `store_result_bodies` (batched,
  inline / spill / no-bucket-degrade, `ON CONFLICT` dedup + recency touch),
  `fetch_result_bodies`, `fetch_full_body` (byte-sliced object read), and
  `sweep_orphan_bodies` (GC).
- `services/provenance_gc.py` (+95, new) — daily singleton mark-sweep with a
  grace window and a non-blocking advisory lock; registered in the app lifespan.
- `app/threads.py` (+170/−1) — two read endpoints: list bodies for a thread and
  fetch one record's full body (`?full=true`), each returning `byte_len` +
  `truncated` + a `verified` flag.
- `database/provenance.py` (+31) — `get_provenance_record` targeted single-row
  lookup for the body endpoint (no whole-thread scan).
- `app/setup.py` (+18) — start/stop the GC service in `lifespan`.

### frontend  (net +52)
- `types/chat.ts` (+49/−9), `pages/ChatAgent/hooks/utils/streamEventHandlers.ts`
  (+14/−2) — a shared `provenanceMcpKey` now discriminates `mcp_tool` sources by
  `(args_fingerprint, result_sha256)` in both the live storage key
  (`provenanceRecordKey`) and the UI dedup key (`provenanceDisplayKey` +
  `countDedupedSources`), so the visible source count reflects it. Formalized
  per-source-type policy: web/SEC/file sources still collapse by URL, but
  `mcp_tool` keeps the sha — identical args returning different data (e.g. live
  market quotes) now render as separate sources instead of collapsing into one.

**What this introduces:** every external data access the agent makes — via
`execute_code` or a script run through `Bash`/`BashOutput` — is now traced, and a
verifier can read the *exact*, hash-consistent bytes the agent saw, stored once
globally, spilled to object storage when large, redacted at rest, without bloating
the SSE stream or the `conversation_responses.sse_events` blob.

## Truncation & verification contract

- `byte_len` is the length of the **stored (post-redaction)** body; a read is
  incomplete exactly when fewer bytes come back than `byte_len`
  (`truncated = byte_len > len(returned)`). A body that redaction shrank below the
  inline cap is stored whole and reads back complete — no false "partial head".
- `mcp_tool` shas/bodies come from the in-sandbox trace, which agent code can
  forge (`MCP_TRACE_FILE` is sandbox-visible), so they're trusted only when the
  held bytes reproduce the claimed hash; on a miss the sha is nulled everywhere
  (SSE event, record, body store). Host-computed shas (web/sec/market) are
  trusted. The 16 MiB host-side trace read cap is the hard bound against an
  oversized agent-authored trace; the codegen body budget is only cooperative.

## Known limitations / follow-ups (not in this PR)

- A backgrounded script's MCP trace is attributed to the `BashOutput` that
  observes completion; if the session is evicted (cap pressure) before any
  `BashOutput` reads it, the trace is dropped — logged, not recovered.
- Empty-args + oversized `mcp_tool` bodies can still collapse in keying.
- `?full=true` reads the whole spilled object into host memory before slicing to
  the 4 MiB read cap.
- Lifting the in-sandbox transport cap (direct sandbox→object-storage upload) is
  deferred.

## Test plan

- `uv run pytest tests/unit/ptc_agent/... tests/unit/server/...` — body store
  (inline/spill/degrade/dedup), per-access capture incl. the bash/bash_output
  trace path and the per-call-not-aggregate `mcp_tool` regression, sha trust gate,
  GC sweep, truncation labelling, threads read endpoints.
- `pnpm vitest run` — Sources keying (storage + display dedup).
- `ruff check` clean on all changed files.
- Migration round-trips `upgrade → downgrade → upgrade`.

## Summary by CodeRabbit

* **New Features**
  * Added provenance result body retrieval APIs (list plus inline/full body) with `truncated` and integrity `verified` indicators.
  * Persist full provenance result bodies content-addressed for later access.
  * Enhanced Bash/BashOutput to surface MCP execution trace details via tool artifacts.
  * Improved MCP provenance dedup/keying to distinguish arguments and time-varying results.
* **Bug Fixes**
  * Added stronger provenance body integrity verification (including redaction/truncation handling).
  * Improved MCP trace harvesting consistency for foreground and background executions.
  * Added periodic garbage collection for orphaned provenance bodies to prevent buildup.

### DIFF
--- a/migrations/versions/015_add_provenance_result_bodies.py
+++ b/migrations/versions/015_add_provenance_result_bodies.py
@@ -1,0 +1,64 @@
+"""Add provenance_result_bodies: a global content-addressed store of raw result bodies.
+
+The 500-char ``result_snippet`` renders the Sources panel but is too truncated
+to verify a response against — the figures a verifier must check (revenue, EPS)
+live in the full result body (e.g. a 126 KB 10-Q), not its metadata head. This
+table stores the raw per-access result body keyed by ``result_sha256`` so a
+post-hoc verifier can read the exact bytes the agent reasoned over. The store is
+global: a static 10-Q fetched by many users across months is stored once. Small
+bodies live inline in ``body_inline``; bodies over the cap keep a head inline and
+spill the full body to object storage (``object_key``). ``byte_len`` is the TRUE
+full length, so ``truncated = byte_len > length(body_inline)``.
+
+No FK from ``provenance_records`` — bodies are shared across rows and GC'd
+independently; ``result_sha256`` is the logical link. Bodies are immutable, so
+writes are ``ON CONFLICT DO NOTHING`` (a dedup hit is a pure no-op). The GC
+mark-sweep deletes rows that no provenance record references and that are older
+than a grace window keyed on ``created_at`` (a body written mid-turn is younger
+than the grace, so it is never reaped before its provenance row commits — no
+per-access ``last_seen_at`` bump is needed). ``idx_provenance_records_sha`` serves
+the NOT EXISTS orphan check + verifier joins; ``idx_provenance_result_bodies_created_at``
+lets the sweep seek old rows without scanning the whole table.
+
+Revision ID: 015
+Revises: 014
+"""
+
+from alembic import op
+
+
+revision = "015"
+down_revision = "014"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute("""
+        CREATE TABLE IF NOT EXISTS provenance_result_bodies (
+            result_sha256 TEXT PRIMARY KEY,
+            body_inline TEXT NOT NULL,
+            object_key TEXT,
+            byte_len BIGINT NOT NULL,
+            content_type TEXT,
+            created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+        )
+    """)
+
+    # Serves the GC mark-sweep NOT EXISTS against provenance_records + verifier joins.
+    op.execute("""
+        CREATE INDEX IF NOT EXISTS idx_provenance_records_sha
+        ON provenance_records(result_sha256)
+    """)
+
+    # Lets the GC sweep seek bodies past the created_at grace window cheaply.
+    op.execute("""
+        CREATE INDEX IF NOT EXISTS idx_provenance_result_bodies_created_at
+        ON provenance_result_bodies(created_at)
+    """)
+
+
+def downgrade() -> None:
+    op.execute("DROP INDEX IF EXISTS idx_provenance_result_bodies_created_at")
+    op.execute("DROP INDEX IF EXISTS idx_provenance_records_sha")
+    op.execute("DROP TABLE IF EXISTS provenance_result_bodies")
--- a/src/ptc_agent/agent/middleware/provenance/middleware.py
+++ b/src/ptc_agent/agent/middleware/provenance/middleware.py
@@ -115,6 +115,15 @@ _MAX_TRACE_ENTRIES = 200
 # event unclamped (the DB writer already coerces a bad value to NULL).
 _TIMESTAMP_MAX_CHARS = 64
 
+# Sanity ceiling for the agent-claimed true result size (the one trace field that
+# otherwise rode through to the SSE event and the BIGINT column unvalidated). It's
+# the TRUE byte length of an MCP result, so it can legitimately exceed the 64 KiB
+# body cap (a truncated body); this only rejects a non-numeric, negative, or absurd
+# value a poisoned trace could forge. Generous — no honest single result approaches
+# it (real maxima are low single-digit MB), but it keeps a fabricated exabyte size
+# out of the record/event.
+_RESULT_SIZE_MAX_BYTES = 256 * 1024 * 1024
+
 # Max concurrent in-flight background body-store writes per middleware instance.
 # A backpressure valve: once this many flushes are outstanding, awrap_tool_call
 # waits for one to finish before scheduling the next, so a burst of source-bearing
@@ -213,6 +222,23 @@ def _clamp_body_bytes(value: Any, limit: int) -> str | None:
     if len(encoded) <= limit:
         return value
     return encoded[:limit].decode("utf-8", errors="ignore")
+
+
+def _coerce_result_size(value: Any) -> int | None:
+    """Trust the agent-authored ``result_size`` only as a bounded non-negative int.
+
+    Every other trace field is clamped on the way in; ``result_size`` rode through
+    raw onto the SSE event and the BIGINT column. On a verified body
+    ``_verify_source_sha`` overwrites it with the real length, so this guards the
+    unverified path (a truncated body, or a forged trace): a non-numeric, negative,
+    or absurd size becomes None rather than being emitted/persisted. ``bool`` is
+    excluded since it is an ``int`` subclass.
+    """
+    if isinstance(value, bool) or not isinstance(value, int):
+        return None
+    if value < 0 or value > _RESULT_SIZE_MAX_BYTES:
+        return None
+    return value
 
 
 def _is_error_result(result: Any) -> bool:
@@ -778,7 +804,7 @@ class ProvenanceMiddleware(AgentMiddleware):
                 args_fingerprint=hash_args(entry.get("args")),
                 args=redact_args(entry.get("args")),
                 result_sha256=_truncate(entry.get("result_sha256"), _SHA_MAX_CHARS),
-                result_size=entry.get("result_size"),
+                result_size=_coerce_result_size(entry.get("result_size")),
                 result_snippet=_truncate(
                     entry.get("result_snippet"), SNIPPET_MAX_CHARS
                 ),
--- a/src/ptc_agent/agent/middleware/provenance/middleware.py
+++ b/src/ptc_agent/agent/middleware/provenance/middleware.py
@@ -14,6 +14,7 @@ from the LangGraph namespace.
 
 from __future__ import annotations
 
+import hashlib
 import logging
 import uuid
 from collections.abc import Awaitable, Callable, Iterable, Iterator
@@ -30,6 +31,10 @@ from ptc_agent.agent.provenance import (
     fingerprint_result,
     hash_args,
     redact_args,
+)
+from ptc_agent.agent.provenance.types import (
+    RESULT_BODY_MAX_BYTES,
+    _canonicalize,
 )
 
 logger = logging.getLogger(__name__)
@@ -113,6 +118,16 @@ _MAX_TRACE_ENTRIES = 200
 _ERROR_CONTENT_PREFIXES = ("[error]", "error:", "failed to ", "exception:", "exception ")
 
 
+def _canonical_body(value: Any) -> str:
+    """Full canonical string of a fingerprinted value — the raw per-access body.
+
+    Must equal the exact string ``fingerprint_result`` hashes for ``value`` so
+    the stored body is hash-consistent with ``result_sha256`` (it is the full
+    body, NOT the snippet-truncated form).
+    """
+    return _canonicalize(value)
+
+
 def _now_iso() -> str:
     return datetime.now(timezone.utc).isoformat()
 
@@ -158,8 +173,16 @@ def _is_agent_infra_path(path: str) -> bool:
     )
 
 
+# Tools whose artifact carries an in-sandbox ``mcp_trace``. All run agent code
+# in the sandbox: ExecuteCode directly, Bash when it runs a script importing the
+# MCP wrappers, and BashOutput which surfaces a backgrounded script's trace when
+# the command finishes. They share the trace extractor, the error-result
+# exemption, and the mcp_trace strip.
+_MCP_TRACE_TOOLS = ("ExecuteCode", "Bash", "BashOutput")
+
+
 def _strip_mcp_trace(result: Any) -> None:
-    """Drop the provenance-only mcp_trace key from an ExecuteCode artifact.
+    """Drop the provenance-only mcp_trace key from an ExecuteCode/Bash artifact.
 
     Mutates the ToolMessage artifact in place so the raw in-sandbox trace never
     propagates outward onto tool_call_result or into persisted sse_events.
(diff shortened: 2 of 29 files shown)


Chen-zexi:
**Addressed the actionable findings from both reviews in `5f4801be`.**

**Fixed**
- `_collect_mcp_trace` wasted round-trip → early-return when `wc -c` reports an empty trace (skips both the read and the `rm` on non-MCP bash commands).
- `_build_trace_env_command` per-invocation `fetch_working_dir()` → reuses the cached `_work_dir` (same value `normalize_path` uses on this path); no Daytona round-trip per bash.
- Evicted bg-session trace JSONL now `rm -f`'d before its path is dropped.
- Agent-controlled trace `timestamp` now clamped (64 chars, type-guarded) like the other trace fields. Note: `_coerce_timestamp` already nulled a bad value at the DB write, so this only stops an oversized string riding the SSE event — "break persisted provenance" wasn't reachable.
- `store_result_bodies(items)` type hint added; duplicate `# 0.` shutdown labels renumbered `0.1/0.2/0.3`.
- Nitpicks: cross-layer `RESULT_BODY_MAX_BYTES` contract test, `BashOutput` middleware-path tests, obviously-fake vault-secret literals.

**Addressed defensively (premise doesn't hold here)**
- "One spill failure aborts the batch" → added per-item `try/except … continue`. For the record, `upload_bytes` catches all exceptions and returns `False`, so `_build_body_row` can't raise on a spill failure — belt-and-suspenders for the best-effort contract.

**Deferred, with rationale**
- bg-launch `except Exception` skipping `CancelledError` → the `_pending:` sentinel exists so `cleanup()` reaps the session at teardown; transient orphan, not a permanent leak. `except BaseException` + `shield` risks a worse partial-cleanup path.
- Per-process `result_body` budget vs the 16 MiB host cap → real but graceful (logged skip; provenance is best-effort); a correct fix needs cross-process shared state. Follow-up.
- Concurrent reuse vs GC object delete → documented the residual window in `sweep_orphan_bodies`: degrades to head-only `full=true`, self-heals on next reuse (unconditional re-upload) → transient degrade, never data loss.
- Empty-body `if not body` → an empty body carries no verification value; keeping the skip.
- Cap-in-SQL for `/provenance/bodies` → `get_provenance_for_thread` is shared with the snippet panel; bounded variant is a fast-follow.

Chen-zexi:
Worked through all four — #1, #2, #4 applied; #3 left as a taste call.

**#1 — `async` with no `await`:** dropped `async` + both `await`s. Correct catch — after the prior round swapped `fetch_working_dir()` for the cached `_work_dir`, the function is pure string work with no I/O left.

**#2 — `full=true` double round-trip:** `fetch_full_body` now takes a preloaded `row=`, and the endpoint passes the `meta` it already fetched from `fetch_result_bodies`, so the second connection/query is gone — only the spilled-object read remains. Added tests asserting the DB query is skipped for both the inline and spilled cases.

**#4 — counter scope:** the premise doesn't hold in production. Each `execute_code` runs in a **fresh interpreter** — both providers run `code_run` as a one-shot `python` (Docker: `python3 <uuid>.py`; Daytona: `process.code_run`, a one-shot run, not a persistent session), so `sys.modules` never survives across runs and `_result_body_emitted_bytes` resets to 0 every execution. The invariant is per-execution, not session-wide accumulation. That said, the comment wording ("imported once per sandbox process") was genuinely misleading — rewrote it to name the fresh-interpreter basis, and tightened the test docstring to match.

**#3 — `_canonicalize` underscore import:** leaving as-is. The coupling is already documented in `_canonical_body`'s docstring, and there's an unrelated `_canonicalize` in `user_data_io.py`, so promoting just this one to a public name trades a convention nit for cross-module name churn. Happy to revisit if you feel strongly.

Chen-zexi:
Landed in `938ca124` (Codex perf review #1 + #3), plus a live end-to-end pass on the PTC + MCP-via-bash path. Per-point against this review:

**2 — `_flush_bodies` awaited before the tool result → fixed.** Exactly the `asyncio.create_task` decoupling suggested here. Body-store writes now run as bounded background tasks (cap 16) scheduled in `awrap_tool_call` and drained in `aafter_agent`, so fast provenance-bearing tools return without blocking on the DB/object write. The dropped "write before next tool call" ordering is safe: the store is content-addressed + idempotent and bodies are read-only after the fact. Verified live — bodies land in `provenance_result_bodies` after the turn closes under the background path.

Same commit also covers #3: extractors now produce the stored body in a single pass via `fingerprint_result_with_body` instead of canonicalizing once to hash and again to store, so the body stays byte-identical to `result_sha256`.

**1 — extra `wc -c` round-trip per foreground bash.** Framing is correct: it's +1 trip for non-MCP bash vs. pre-PR, not −1. Keeping it — the `wc -c` probe is the 1-trip gate that avoids the read+rm (2 trips) when no trace file exists, and it's the cost of the new capability (bash-invoked MCP calls are now traceable). Folding the probe into the executed command itself (zero extra trips) is a real win but a separate perf item, deliberately deferred this round. Skipping the `.py`/`python` heuristic — agree it's not worth the complexity, and it would miss MCP calls from non-obvious commands.

**3 — lazy agent→server import.** Accurate and prod-safe (guarded by `except Exception` + warning; in standalone use it degrades to a logged warning, not a crash). Leaving the import as-is for now; the forward-declared Protocol/ABC seam is worthwhile but a larger refactor, and a fair follow-up.

Live E2E (real backend + Daytona sandbox + Postgres + COS, OSS mode), focused on the bash path:
- **MCP-via-bash:** a `python` script calling 3 MCP tools through `tools.*` → each call's body stored individually (not the aggregate bash stdout — the original regression), `sha256(body_inline) == result_sha256`, duplicate call dedups to a single body row.
- **>64 KiB MCP body:** `result_sha256` + body correctly nulled by the SHA trust gate (the 64 KiB head can't verify against the full-result sha), `result_size` preserved on the record.
- **web_search fan-out:** 10 per-item bodies, all hash-consistent.
- **Object spill:** existing >64 KiB bodies round-trip from COS — full object = `byte_len` bytes and hashes to `result_sha256`.
- **Failed `web_fetch`** (upstream 400) → correctly produced no provenance row.
- `sse_events` carries no `result_body` on any path.

Chen-zexi:
Addressed in `6f676e29`. Verified each finding against current code; fixed the valid ones, kept changes minimal.

**B1 — full-table scan in `get_provenance_bodies` → fixed.** New `get_provenance_body_refs(conn, thread_id, limit)` selects only `provenance_record_id, result_sha256` with `WHERE result_sha256 IS NOT NULL ... LIMIT %s`, and the endpoint runs it + `fetch_result_bodies` on one shared connection — O(limit) instead of O(all records), no more shipping every row's `args` JSON to discard most. Two deltas from the suggested patch: kept the existing `ORDER BY turn_index … ` rather than `created_at DESC` (the latter changes *which* records return when capped), and used `LIMIT limit+1` so the `capped` flag survives the SQL cap. Live-checked against Postgres: `limit=3` on a 10-body thread → 3 records, `capped=true`.

**Outside-diff: `result_size` trust boundary → fixed.** It was the one trace field passed through raw onto the SSE event + the BIGINT column. Now coerced to a bounded non-negative int (`bool` excluded, ≤256 MiB sanity ceiling, else `None`) at the trace-assembly site, alongside the existing sha/snippet/body/timestamp clamps. One deliberate deviation: I **sanitize** rather than drop-on-verify-fail, because `result_size` is intentionally preserved when a >64 KiB body fails the SHA gate (it's the only signal that a large truncated result existed). Dropping it on every verify failure would erase that; sanitizing the magnitude fixes the actual gap (malformed/oversized values escaping) without the regression.

**N2 — private `_GC_GRACE_DAYS` imported across modules → fixed.** Renamed to `GC_GRACE_DAYS`; it's genuinely a shared constant (the GC service and `sweep_orphan_bodies` both source their grace from it), so the underscore was the misleading part. Checked for other cases: the only cross-module private import in this PR's surface was this one (other `_`-prefixed imports elsewhere in the tree are pre-existing and unrelated).

**N1 — duplicate `RESULT_BODY_MAX_BYTES` → skipped (by design).** The re-declaration is deliberate, to keep the server-side module out of the agent package's import graph (the comment flags it). Low divergence risk on a stable constant; if you want a guard, a one-line equality test is cheaper than a shared-module refactor.

ruff clean; 254 unit tests pass (new `result_size` coercion test + updated body-list endpoint test). Backend reloaded clean on the live container.

Chen-zexi:
Verified each finding against current code on `feat/provenance-result-full`. Net: merging as-is — B1 is mis-scoped and incompatible with our Alembic env, N1 is a low-priority cold-path micro-opt, and N2 is already guarded by existing tests.

**B1 — `CREATE INDEX` without `CONCURRENTLY`: declining for this migration.**
- Only one of the two indexes is on a pre-existing table. `idx_provenance_result_bodies_created_at` is on `provenance_result_bodies`, **created in this same migration** (line 38) — empty table, no concurrent DML, so `CONCURRENTLY` is moot there. Only `idx_provenance_records_sha` touches a live table.
- `provenance_records` was introduced in migration `013` on this same feature line, so it's small; the SHARE-lock window for the build is brief.
- The drop-in fix would *fail* in our setup: `migrations/env.py:78` wraps every migration in `context.begin_transaction()`, and `CREATE INDEX CONCURRENTLY` cannot run inside a transaction block (PG `25001`). Making it work needs `op.execute("COMMIT")` injection (risks a half-applied migration if the concurrent build then errors) or an env-wide non-transactional change — both larger and riskier than the lock they'd avoid. No migration in this repo uses `CONCURRENTLY`; every `CREATE INDEX` (including all of `001`) is plain and on same-migration tables.
- If a zero-write-stall guarantee against a large `provenance_records` ever becomes a requirement, that's a deliberate scoped follow-up (split the index op + mark the migration non-transactional), not a one-liner.

**N1 — two connections in `get_provenance_record_body`: confirmed, skipping.** Low-QPS manual verifier endpoint; warm-pool acquisition is cheap; and the costly path is already optimized — `fetch_full_body(sha, row=meta)` (`threads.py:1511`) reuses the fetched `meta` to skip a third connection. Merging the remaining two would mean threading a `conn` param through `get_provenance_record`; not worth it on this path.

**N2 — duplicate `RESULT_BODY_MAX_BYTES` not machine-checked: already guarded.** Two cross-module equality tests already pin the copies equal, one from each direction: `test_provenance_bodies.py::test_max_bytes_matches_canonical_agent_constant` (`assert RESULT_BODY_MAX_BYTES == AGENT_MAX_BYTES`) and `test_middleware_result_body.py::test_result_body_cap_matches_across_layers` (`assert RESULT_BODY_MAX_BYTES == SERVER_CAP`) — the one-line drift guard suggested here, already present from both sides.

PR #289, opened 2026-06-27
fix(report-back): reliable, live-streaming flash↔PTC report-back

Chen-zexi:
## Summary

Makes the flash→PTC **report-back** path work end-to-end for the single-job case.
When a flash turn dispatches a background PTC research job via the secretary
`ptc_agent` tool, the job's summary should stream back into the flash thread live
on completion. Three defects broke that:

**Dispatch fix** (`52b03d9c`) — `ptc_agent` continuation always failed with "thread
not found". The `is_continuation` branch compared `thread.get("user_id")`, but
`conversation_threads` has no `user_id` column (ownership lives on
`workspaces.user_id`), so the value was always `None`. Now reuses
`_verify_thread_owner`, which JOINs `workspaces` to resolve the real owner.

**Report-back fix** (`af35a6c2`) — the flash thread was not reliably notified when
the PTC job finished. `_flash_report_back` cleared the Redis watch state before
publishing the wake (and even on a failed POST), so a client that missed the live
pub/sub never reopened the watch on reload. The watch (`flash_watch` + `ptc_origin`)
now clears only when the report-back flash run reaches a terminal state: success via
the completion hook, fail/cancel via `BackgroundTaskManager`. The wake payload and a
durable `flash_rb_run` pointer both carry the `run_id`, and `/status` returns
`report_back_run_id`, so clients attach to the exact run instead of inferring it.
`report_back_ptc_thread_id` is stripped for non-internal callers (anti-spoof).

**Streaming fix** (`998ab08d`) — in-session the summary rendered as an empty spinner
and only appeared after a refresh. `watchThread` reacted on a partial SSE buffer and
regex-parsed possibly-split JSON, losing the `run_id` so the one-shot watch closed
with nothing to attach. Now splits the buffer on complete frames, attaches to the
named run from the wake, and keys the watch to the flash thread so it survives a jump
into the live PTC thread and replays on return. Stream ownership is tracked via an
`acquire/releaseStreamOwnership` invariant so the flash report-back and a live PTC
stream no longer fight over the single visible slot.

## Overview

| Category | Files | +LOC | -LOC |
|---|---:|---:|---:|
| Modified (production) | 9 | 579 | 129 |
| New (production) | 0 | 0 | 0 |
| Tests (4 new + 2 modified, excluded from net) | 6 | 1295 | 0 |
| **Net (excl. tests)** | **9** | **579** | **129** |

**By module**

- **Backend / server** (`src/server/`, modified)
  - `handlers/chat/ptc_workflow.py` (+152/-54) — durable `flash_rb_run` pointer,
    bounded report-back retry, terminal-state clear, run_id in the wake.
  - `services/background_task_manager.py` (+41/-3) — `_clear_report_back_watch` wired
    into `_mark_failed` / `_mark_cancelled` so failed/cancelled runs clear the watch.
  - `handlers/chat/flash_workflow.py` (+45) — clear-on-success hook; threads
    `report_back_ptc_thread_id` into task metadata.
  - `app/threads.py` (+18/-9) — PTC-crash-path clear; strip `report_back_ptc_thread_id`
    for non-internal callers.
  - `handlers/workflow_handler.py` (+10/-1) — `/status` returns `report_back_run_id`.
  - `models/chat.py` (+9) — `report_back_ptc_thread_id` field on `ChatRequest`.
- **Backend / tools** (`src/tools/secretary/tools.py`, +5/-2, modified) — continuation
  uses `_verify_thread_owner` instead of the always-`None` `thread.get("user_id")`.
- **Frontend / ChatAgent** (`web/src/pages/ChatAgent/`, modified)
  - `hooks/useChatMessages.ts` (+270/-50) — keyed nav-surviving report-back watch,
    stream-ownership invariant, epoch guard, bounded missed-wake poll.
  - `utils/api.ts` (+29/-10) — `watchThread` frame-splitting; `reconnectToStream`
    attaches to a named run with cursor reset.

**What this introduces:** a single background PTC job dispatched from a flash thread
now reports its result back and streams live into that flash thread, surviving
navigation to the PTC thread and surviving a reload, instead of erroring on dispatch,
silently dropping the notification, or rendering only after a manual refresh.

## Diff Size
- Total: 15 files changed, 1874 insertions(+), 129 deletions(-)
- Net (excl. tests): 9 files changed, 579 insertions(+), 129 deletions(-)

## Test Coverage

New/extended unit coverage:
- `tests/unit/tools/secretary/test_ptc_agent_continuation.py` (new) — continuation
  owner-verification path.
- `tests/unit/server/handlers/chat/test_flash_report_back.py` (new) — report-back
  retry/clear lifecycle, terminal-state clear, run_id in wake/pointer.
- `tests/unit/server/services/test_background_task_manager.py` — `_clear_report_back_watch`
  via terminal handlers.
- `web/.../__tests__/useChatMessages.reportBack.test.ts` (new) — live attach,
  missed-wake reload via `/status`, nav supersede + replay, no-run no-attach.
- `web/.../__tests__/useChatMessages.reconnectNav.test.ts` (new) — reconnect/navigation
  ownership.
- `web/.../utils/__tests__/api.test.ts` — `watchThread` frame-splitting.

**Known gaps (deferred, none blocking):** a direct assertion that
`report_back_ptc_thread_id` is stripped for non-internal callers (security control),
the `_mark_failed`/`_mark_cancelled` → clear wiring as an isolated case, and the
`REPORT_BACK_MAX_POLLS` give-up branch (needs fake timers).

## Pre-Landing Review

Reviewed via `/review` (no hard blockers). One auto-fix applied: `run_id?: string`
→ `string | null` in `useChatMessages.ts` (backend returns `null`). Deferred nits
(non-blocking): per-attempt `aiohttp.ClientSession` in `_flash_report_back`, a
defense-in-depth origin match in `clear_flash_report_back`, and a duplicate
`_FLASH_RB_RUN_TTL` constant. No SQL/data-safety or trust-boundary issues; the
anti-spoof strip is correct (POST host fixed to `GINLIXFLOW_BASE_URL`, only a
server-generated UUID interpolated, no SSRF).

## Scope

Single-job report-back only. The **concurrent multi-dispatch** case (2+ background
jobs from one flash thread finishing close together) is a known follow-up, designed
but not built. It touches the same files and stacks on this change.

## Test plan
- [x] Backend unit, broad (`tests/unit/server` + `tests/unit/tools`) — 2600 passed
- [x] Backend unit, targeted (secretary + chat handlers + BTM) — 139 passed
- [x] Frontend unit, targeted (ChatAgent hooks + api) — 157 passed
- [x] `ruff check src/` clean; `pnpm lint` 0 errors
- [x] E2E (`/e2e-chat`, live worktree container, oss-mode) — dispatch, report-back,
      and live-stream paths exercised against the running backend; report-back
      dispatched and streamed live (incremental chunks, no refresh)

## Summary by CodeRabbit

* **New Features**
  * Added support for reporting a specific report-back run identifier in workflow status to enable more accurate reconnection.
* **Bug Fixes**
  * Improved chat report-back reconnect behavior so users resume the correct stream when navigating between threads.
  * Enhanced flash/PTC report-back cleanup on completion, cancellation, and failure to reduce stale “pending” states.
  * Strengthened chat continuation by denying continuation attempts for threads that aren’t owned, returning clearer “thread not found” style errors.

### DIFF
--- a/src/server/app/threads.py
+++ b/src/server/app/threads.py
@@ -90,7 +90,11 @@ def _track_task(task: asyncio.Task) -> None:
 
 
 async def _consume_background_gen(
-    gen, label: str, thread_id: str, run_id: str
+    gen,
+    label: str,
+    thread_id: str,
+    run_id: str,
+    report_back_ptc_thread_id: str | None = None,
 ) -> bool:
     """Drain an async generator in the background, cleaning up Redis on failure."""
     _ok = True
@@ -113,14 +117,22 @@ async def _consume_background_gen(
 
             cache = get_cache_client()
             if cache.enabled and cache.client:
-                # Only the PTC generator crashing reaches this with a real
-                # ptc_origin: a flash-dispatch crash looks up
-                # ptc_origin:{flash_thread_id} (wrong key) -> no-op, so the
-                # report-back keys survive for reload recovery.
-                origin = await cache.get(f"ptc_origin:{thread_id}")
+                # Resolve the PTC thread whose report-back watch this crash should
+                # tear down. The origin is keyed by the *PTC* thread id:
+                #   - PTC_DISPATCH crash: thread_id IS the ptc thread -> direct hit,
+                #     clear (the watched run just died).
+                #   - report-back run crash: report_back_ptc_thread_id names the
+                #     completed PTC -> clear, else a never-finishing report-back
+                #     leaves /status reporting a stale pending run until TTL.
+                #   - ordinary flash-dispatch crash: no report_back id and thread_id
+                #     is the flash thread -> ptc_origin:{flash_tid} misses -> no-op,
+                #     so a still-running dispatched PTC's keys survive for reload
+                #     recovery.
+                ptc_thread_id = report_back_ptc_thread_id or thread_id
+                origin = await cache.get(f"ptc_origin:{ptc_thread_id}")
                 if origin:
                     flash_tid = origin.get("flash_thread_id")
-                    await clear_flash_report_back(cache, thread_id, flash_tid)
+                    await clear_flash_report_back(cache, ptc_thread_id, flash_tid)
                     if flash_tid:
                         await cache.client.publish(
                             f"thread:wake:{flash_tid}",
@@ -637,7 +649,13 @@ async def _handle_send_message(
                 await manager.pre_register(thread_id, run_id)
             _track_task(asyncio.create_task(
                 observe_background_chat_turn(
-                    _consume_background_gen(flash_gen, "FLASH_DISPATCH", thread_id, run_id),
+                    _consume_background_gen(
+                        flash_gen,
+                        "FLASH_DISPATCH",
+                        thread_id,
+                        run_id,
+                        report_back_ptc_thread_id=request.report_back_ptc_thread_id,
+                    ),
                     mode="flash",
                     model=_model,
                     user_id=user_id,
--- a/src/server/app/threads.py
+++ b/src/server/app/threads.py
@@ -107,19 +107,21 @@ async def _consume_background_gen(
         )
         try:
             from src.utils.cache.redis_cache import get_cache_client
+            from src.server.handlers.chat.ptc_workflow import (
+                clear_flash_report_back,
+            )
 
             cache = get_cache_client()
             if cache.enabled and cache.client:
+                # Only the PTC generator crashing reaches this with a real
+                # ptc_origin: a flash-dispatch crash looks up
+                # ptc_origin:{flash_thread_id} (wrong key) -> no-op, so the
+                # report-back keys survive for reload recovery.
                 origin = await cache.get(f"ptc_origin:{thread_id}")
                 if origin:
                     flash_tid = origin.get("flash_thread_id")
-                    await cache.delete(f"ptc_origin:{thread_id}")
+                    await clear_flash_report_back(cache, thread_id, flash_tid)
                     if flash_tid:
-                        watch_key = f"flash_watch:{flash_tid}"
-                        await cache.client.srem(watch_key, thread_id)
-                        remaining = await cache.client.scard(watch_key)
-                        if remaining == 0:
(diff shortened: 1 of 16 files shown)


Chen-zexi:
Verified both findings against the current branch.

**CodeRabbit — "Publish `run_id` in the failure wake" (`threads.py:136-170`) — not applying.**
The premise that the flash watcher's only `run_id` source is this wake or `/status` is incorrect. `_flash_report_back` (`ptc_workflow.py:276-294`) always publishes a wake carrying `run_id` *and* persists the `flash_rb_run` pointer on a successful dispatch — which is the precondition for `_consume_background_gen` running at all. The frontend learns the id there (`reconcile('wake', wakeRunId)` → `reportBackRunIdRef`) and attaches to `workflow:stream:{flash_tid}:{rb_run_id}`, the exact key the terminal error SSE is `xadd`'d to (`threads.py:168-171`). So the error SSE is already reachable; this wake is a redundant secondary signal.

The proposed blanket `"run_id": run_id` is also unsafe on the shared PTC_DISPATCH path: there `run_id` is the PTC run and its error SSE lives on `workflow:stream:{ptc_tid}:{ptc_run}`, but the wake targets the *flash* thread (`thread:wake:{flash_tid}`, resolved from `ptc_origin`). The flash reconcile would then attach to `workflow:stream:{flash_tid}:{ptc_run}` — a nonexistent key — turning today's harmless no-op into a spurious empty-stream reconnect. Skipping.

**claude[bot] — `setInterval` fire-and-forget nit (`useChatMessages.ts:2544`) — acknowledged, non-blocking, leaving as-is.**
`reconnectToStream` wraps all awaited work in a top-level `try/catch/finally` that swallows every path, including its own `'Reconnection stream disconnected'` throw (caught at the same level), and `getWorkflowStatus` inside `reconcile` is independently guarded. So the interval callback cannot reject today; a `.catch()` would be defense-in-depth only, with no behavioral effect.

PR #291, opened 2026-06-28
fix(chat): recover the chat stream after iOS Safari backgrounds the tab

Chen-zexi:
## Problem

On iPhone Safari, backgrounding the browser during an active chat turn and returning hangs the chat and shows a dead-end **"failed to load"** banner. iOS Safari freezes a backgrounded tab's JS and tears down its network connections; the chat consumes a long-lived SSE stream over raw `fetch()` + `ReadableStream`, so the frozen-then-killed socket breaks the stream. The app already has full reconnect/resume infrastructure — a string-match bug just stopped Safari from ever reaching it.

## Root cause

The SSE reader loop in `streamFetch()` only classified a thrown error as a recoverable disconnect when `error.name === 'TypeError' && message.includes('network')`. iOS Safari's background-kill rejects the read with `TypeError: Load failed` — no `"network"` substring — so it fell through to `throw`, surfacing the error banner and never triggering the existing 5-retry reconnect.

Secondary gaps: nothing kicked recovery if the frozen `read()` didn't reject promptly on resume (the hang); the reconnect path keyed off the `threadId` prop, which is still `'__default__'` during a brand-new chat's first-answer window; and a refetch-on-focus storm could 401 before Supabase's frozen auto-refresh timer caught up.

## What changed (A + B + C)

- **A — Classify Safari transport drops as reconnects.** Any `TypeError` escaping the SSE read loop is now treated as a transport drop and routed into `attemptReconnectAfterDisconnect`, reusing the existing resume machinery (`getWorkflowStatus().can_reconnect` → `Last-Event-ID` resume, or history reload if the run already finished).
- **B — Proactive foreground reconnect + visible state.** A return-to-foreground handler aborts the likely-dead reader and re-kicks reconnect at every main-stream result site (send, reconnect, HITL, checkpoint, demoted steering) — gated so a background abort is never mistaken for a user stop. The reconnect path keys off the **latched `threadIdRef`** instead of the `threadId` prop, so a first-turn chat still on the `'__default__'` route recovers. A "Reconnecting…" indicator surfaces the resume.
  - **Scoped to genuine suspends only:** the reconnect arms on `pagehide`/`freeze` (the OS actually froze the tab and killed the socket), not on a bare `visibilitychange`. A desktop alt-tab fires neither and keeps the socket alive, so non-suspended users see **no behavior change** — no needless teardown, no "Reconnecting…" flash on refocus.
- **C — Token-refresh resilience.** The axios response interceptor does a single-shot refresh-and-retry on 401 (`config._retry` guards against loops); `AuthContext` registers/clears the refresher; `getAuthHeaders` force-refreshes an expired/near-expiry token before opening the SSE stream. Local-dev (no Supabase) registers none and behaves exactly as before.

## Overview

**Totals** — 9 files, **+713 / −24**

| | Files | Net (excl. tests) |
|---|---|---|
| Source (modified) | 5 | **+232 / −23** |
| Tests (1 new, 3 modified) | 4 | +481 / −1 |

**By module (source):**

| File | Δ | What |
|---|---|---|
| `ChatAgent/hooks/useChatMessages.ts` | +160 / −15 | suspend-gated foreground-reconnect effect, re-kick at all result sites, `__default__`-aware reconnect keying, demote-path re-kick |
| `ChatAgent/utils/api.ts` | +29 / −4 | broaden read-loop `TypeError` → disconnect; force-refresh SSE auth token |
| `api/client.ts` | +29 / −2 | `setTokenRefresher` + single-shot 401 refresh-and-retry |
| `ChatAgent/components/ChatView.tsx` | +9 / −1 | "Reconnecting…" indicator |
| `contexts/AuthContext.tsx` | +5 / −1 | register refresher on login, clear on logout |

**What this introduces:** a small foreground-reconnect lifecycle surface (one effect + per-result-site re-kicks) armed only on genuine tab suspend, latched-thread reconnect keying for the first-answer window, and a reactive 401 refresh interceptor on the REST path. No architectural change — all resume logic is reused, not reinvented.

## Testing

- New `useChatMessages.visibilityReconnect.test.ts` (7 tests): foreground re-kick after a real suspend on `visibilitychange`/`pageshow`; **no reconnect on a bare tab-switch with no suspend** (desktop no-regression guard); inert when no stream is active / tab still hidden; demoted-steering re-kick; and the **first-turn `'__default__'` reconnect** (renders on `'__default__'`, latches a real server id, asserts reconnect targets the real id — verified fail-before/pass-after).
- `api.test.ts`: `TypeError: Load failed` and `The network connection was lost.` both resolve `{ disconnected: true }` instead of throwing.
- `client.test.ts`: 401→200 single refresh+retry succeeds; 401→401 rejects (no loop); no refresher ⇒ rejects (local-dev parity).
- Full suite: **182 files, 2019 passed**. `tsc --noEmit` clean. Lint clean on changed files. Each of the 4 commits is green in isolation — bisectable.

## Still pending / follow-ups

- **True-device verification** (gold standard): physical iPhone Safari (or iOS Simulator) — background a long PTC turn ~30–60s, return, expect a brief "Reconnecting…" then resume/complete with no error banner. Desktop Safari only partially reproduces (keeps the socket alive — now intentionally a no-op).
- **Fast-follow #2 — connect-time TypeError:** the initial-fetch catch in `api.ts` still throws on a *reconnect's* opening fetch (Prong A is scoped to the mid-read catch).
- **Fast-follow #3 — SSE-path 401 retry:** raw-`fetch()` SSE streams aren't covered by the axios 401 interceptor; `getAuthHeaders` force-refresh mitigates proactively, but there's no reactive refresh-and-retry if an SSE reconnect itself 401s.
- **iOS edge case:** if iOS freezes a tab without firing `pagehide`/`freeze` (rare), the proactive handler won't arm — Prong A still recovers it once the frozen `read()` eventually rejects.
- **i18n:** add `chat.reconnecting` to `en-US.json` / `zh-CN.json` (currently an inline English fallback).

## Summary by CodeRabbit

* **New Features**
  * Added a visible “Reconnecting…” status indicator in the chat UI while the app recovers.

* **Bug Fixes**
  * Improved chat stream recovery after tab switching and app backgrounding, preventing stuck or lost streams.
  * Enhanced request retry behavior: failed requests refresh the token once on unauthorized responses, then retry safely without looping.
  * Improved live-stream reliability by refreshing near-expiry sessions and treating mobile transport interruptions as reconnectable events.

### DIFF
--- a/web/src/api/__tests__/client.test.ts
+++ b/web/src/api/__tests__/client.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
-import { setTokenGetter } from '../client';
+import { api, setTokenGetter, setTokenRefresher } from '../client';
 
 interface InterceptorHandler<T = unknown> {
   fulfilled: (value: T) => T | Promise<T>;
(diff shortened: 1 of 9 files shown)


Chen-zexi:
**#1 (broadened `TypeError` catch) — won't fix; the concern doesn't hold.** `processEvent` runs as `onEvent(d)` *inside* `processLine`'s existing try/catch (`api.ts:415-422`), which this PR does not touch. A `TypeError` thrown by event processing is caught and logged there as an SSE parse error, so it never propagates to the outer `reader.read()` catch at `api.ts:447` and is never routed into the reconnect path — before or after this PR. The "previously surfaced as an error banner" premise is also inaccurate: those exceptions were already swallowed by that same guard. The broadened catch only ever sees `reader.read()`/`decoder.decode` rejections (genuine transport drops), so the scope is correct and the inner-try refactor / diagnostic note aren't needed.

**#2 (vestigial `threadId` dep) — valid, fixed in d9fd234e.** The handlers read `threadIdRef.current`, not the `threadId` prop, so the dep only re-registered the lifecycle listeners on every thread navigation. Dropped it; `isLoading` stays (the one closure-read value). tsc/lint clean, suite green (2019).

PR #298, opened 2026-07-02
Feature/ace tool selection

lastfreeyang-lgtm:
# feat(ace): Integrate Agent Context Engine (ACE) for Sparse Tool Selection

## Description
This PR integrates the **Agent Context Engine (ACE)** into the LangAlpha Programmatic Tool Calling (PTC) pipeline. It introduces dynamic **Sparse Tool Selection** to significantly reduce token overhead and optimize prompt caching.

### Background & Problem
Currently, LangAlpha injects the **entire schema summary** of all active MCP servers (over 10+ servers, dozens of tools) into the system prompt on every single chat turn. This causes:
1. **Token Bloat**: A permanent overhead of ~20,000+ input tokens per turn.
2. **Prompt Cache Inefficiency**: The static system instructions are constantly invalidated by dynamic state changes, causing high costs.

### Solution: ACE Sparse Selection
We introduce a 4-phase integration that dynamically selects only the relevant tools for the user query, tracing dependencies using a Directed Acyclic Graph (DAG) built at server startup.

```mermaid
flowchart TD
    A["User Query"] --> B["SparseToolContextMiddleware"]
    B --> C["QueryToolMatcher (v2 LLM)"]
    C --> D{"Classification LLM\n(Configured or Main fallback)"}
    D -- "Success (JSON)" --> E["Server Names\n['price_data', 'fundamentals']"]
    D -- "Failure (API Error/Timeout)" --> F["All Tools Fallback"]
    E --> G["ACE: MCP Schema Graph (BFS)"]
    F --> H["Full Tool Summary"]
    G --> I["Selected Tool Summary"]
    I --> J["Injected after Prompt Cache Breakpoint"]
    H --> J
    J --> K["Main Agent LLM (Claude)"]

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added Agent Context Engine (ACE) configuration with query-based sparse tool selection, category-to-server routing, and optional classification guidance.
  * Agents can now inject a sparse, request-specific tool context into system prompts to reduce tokens.
* **Bug Fixes**
  * Improved robustness when tool selection/classification fails by safely falling back to the full tool set.
* **Tests**
  * Added unit tests for ACE components and a token-savings integration benchmark.
* **Chores**
  * Made manifest/model loading use explicit UTF-8 encoding; updated a CI boundary test expectation and path normalization for consistency.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


Chen-zexi:
Hey @lastfreeyang-lgtm, thanks for proposing this, but I'm going to close it. A few notes on why:

**On the premises:**

> Currently, LangAlpha injects the entire schema summary of all active MCP servers (over 10+ servers, dozens of tools)

This isn't the case. MCP tools use progressive disclosure. By default each server is injected in `summary` mode: a short header, the module path, the tool count, an import hint, and a pointer to `tools/docs/{server}/{tool}.md`. The actual tool schemas and signatures are not in the prompt; the agent reads them on demand from those docs in the sandbox. Verbosity is configurable per server via `tool_exposure_mode`. So there is no ~20k token schema dump per turn to reclaim.

> The static system instructions are constantly invalidated by dynamic state changes

This is also not the case, and it is the part your change actually makes worse. Today the tool summary is static: it is computed once per session, kept stable, and lives inside the prompt cache prefix (the cache breakpoint sits on the static prompt, and dynamic content like time, profile, and agent.md is appended after it so it cannot invalidate the cached prefix). The summary only changes, and therefore only invalidates the cache, in special cases such as a new tool being registered or the MCP config changing. Those are rare.

Your change makes the tool summary depend on the user's query, so it changes on every turn, and in this implementation on every model step. That is exactly the "constantly invalidated by dynamic state" pattern you describe as the problem, except it does not exist today, and your change would be the single biggest source of it. A tool summary that depends on the query is fundamentally incompatible with prefix caching: in its current placement the sparse block sits after the cache breakpoint, so it is never cached and gets billed again on every step; and the only way to make it cacheable would be to move it into the cached region, where its variation from turn to turn would then invalidate the cache every turn. Either way you lose the caching you have today.

**On the approach:**

We won't put a middleware on the critical path that makes an LLM call to route tools. In this implementation it runs on every model step (once per iteration of the agent loop, not once per turn), with no timeout and no memoization, so a slow or failing classifier stalls every step of the run. The latency and failure surface don't justify the gain, especially since the summary it is trying to trim is already small and already cached.

**One correctness issue to call out:** the injected block tells the model "Only the following MCP tools are loaded for this turn. Do NOT attempt to use tools not listed here." but nothing is actually unloaded. Every wrapper module is still in the sandbox and importable, and the full summary is still in the same prompt. So it is an unenforced, contradictory instruction: when the classifier picks the wrong subset, we are telling the model to avoid tools it needs and that are still available. That is a real correctness risk with no offsetting benefit.

Net effect: there is no benefit here to weigh against any of the above. As written it does not cut tokens, it adds them, plus an extra main model classification call on every step. And the automated reviews already on this PR confirmed the same implementation bugs: the `ace` config is never read so the feature runs on hardcoded defaults (classification silently falls back to the main model), the Redis graph cache never persists, and the full summary is still injected on top of the sparse one. So I am closing rather than iterating on it.

lastfreeyang-lgtm:
I analyzed it incorrectly. Thank you for your reply.

PR #301, opened 2026-07-03
feat(workspaces): resource tiers, always-on, duplicate/rename

Chen-zexi:
### Summary
Adds end-to-end workspace management — change resource tier, enable/disable always-on, duplicate, and rename — with per-plan entitlement checks (a no-op in OSS mode) and the gallery UI to drive it.

- **Sandbox:** per-tier snapshot sizing + auto-stop interval control down through the provider (autostop is capability-gated, so providers without it no-op).
- **Server:** new endpoints + `resource_tier`/`is_always_on` columns (migration 016); running-workspace tier changes recreate via backup → destroy → recreate → restore with files preserved through the DB; the idle reaper reclaims lapsed always-on and lazily reclaims elevated tiers on reprovision.
- **Web:** per-card action menu, tier/always-on badges, four dialogs, shared mutation hook surfacing structured quota errors.

### Overview
| Category | Files | +LOC | −LOC |
|---|--:|--:|--:|
| Modified | 19 | ~1627 | ~347 |
| New | 7 | ~1049 | 0 |
| Tests (excl.) | 7 | ~2324 | ~11 |
| **Net (excl. tests)** | **26** | **+2676** | **−347** |

**By module:** sandbox/provider tier+autostop plumbing (`src/ptc_agent/*`); server workspace-management service + API + migration (`src/server/*`, `migrations/016`); workspace management UI (`web/src/pages/ChatAgent/*`); Daytona SDK bump (`uv.lock`).

**What this introduces:** workspaces go from a single fixed-size, always-idle-reaped sandbox to a managed, tiered, optionally-always-on resource with duplicate/rename and per-plan quota checks.

### Known limitations (verified, deferred by design)
- Count-quota check is TOCTOU across a user's workspaces — concurrent upgrades can both pass; a full fix needs an atomic server-side reserve.
- Duplicate uses best-effort backup (source stays intact → a transient failure yields a slightly-stale copy, not data loss).
- Restore-after-recreate is best-effort but backed by strict pre-teardown backup persisted to the DB, so a partial restore is degraded-not-lost and re-restorable.
- Operator-defined tiers outside `standard|performance|max` are ungated for internal callers only; the public API is Literal-bound.

### Test plan
- [x] Backend unit: 5206 passed, 1 xfailed
- [x] Frontend: 2100 passed
- [x] ruff + `uv lock --check` clean
- [x] Live E2E matrix (create/rename/duplicate/spec/always-on/quota checks/reclaim/file-persistence) passed 2026-07-01

## Summary by CodeRabbit

* **New Features**
  * Added workspace resource tiers (standard/performance/max) with quota-aware spec updates and tier-based sandbox sizing.
  * Added “always-on” support (24/7) with creation-time auto-stop control and syncing for running workspaces.
  * Added new workspace APIs: workspace quota, spec change, always-on toggle, and duplication.
  * Updated the UI with tier selection, always-on confirmation, and duplication flows.
* **Bug Fixes**
  * Improved/standardized error handling and workspace action status responses.
* **Tests**
  * Expanded unit tests for tier snapshot sizing, quota/entitlement gates, and always-on reconciliation.


Chen-zexi:
All three non-blocking items reviewed; two fixed in 1eb7e96d, one accepted.

1. **Frontend test gap (ChangeSpecDialog / useWorkspaceMutation)** — accepted, consistent with the coverage line drawn in the pre-PR review round; the hook's quota-invalidation and rollback behavior is exercised indirectly through the gallery flow.
2. **Silent clamp** — fixed: `_clamp_tiers_to_ceiling` now logs a warning per clamped resource, and `validate_default=True` makes the validator run on the defaults too.
3. **`X-RateLimit-Limit: ""`** — fixed: the header is omitted when the limit is unknown. The three pre-existing 429 sites in the module keep their existing pattern; normalizing those is out of scope here.

Chen-zexi:
Findings triaged; fixes landed in 1eb7e96d (backend) and c7ac2725 (web).

**Fixed**
- `resource_tiers` field validator now has `validate_default=True`. No behavior change today — the built-in defaults are within the ceiling — but the gap was real.
- Duplicate flow now invalidates the workspace quota query. Confirmed real: the copy inherits the source's tier, so it can consume a per-tier quota slot the UI didn't refresh.
- Fail-open (`None`) scope answers get a 15s negative TTL instead of the full 300s, mirroring `_fetch_platform_membership`.
- Tier clamping logs a warning per clamped resource.
- `setWorkspaceSpec` takes the existing `ResourceTier` union instead of `string`.

**Working as designed**
- Tier-reclaim inside `_restart_workspace` on the lazy path: the `stopped → starting` DB claim exists precisely to serialize sandbox rebuilds across workers, and the reclaim path mirrors the existing `SandboxGoneError` recovery, which also runs under that claim. Hoisting the destroy/recreate out of the claim would reintroduce the double-provision race the claim prevents. Waiters queue outside the in-process lock (see the comment at the `_claim_and_restart` callsite), so the cost is bounded to concurrent starters of the same workspace — which must wait for the rebuild regardless.

**Deferred**
- Client timeouts on the new mutation calls: a spec change takes ~14s and duplicate provisions a sandbox eagerly (minutes for heavy workspaces). A timeout tight enough to matter would abort the client while the server completes, reporting failure for an operation that succeeded. If added, it needs generous bounds (~120–180s); left out for now.
- Duplicate-dialog hook reuse: agreed low value, per your own rating.

Chen-zexi:
Confirmed — the blocking finding was real, and my earlier works-as-designed reply defended the wrong lock: holding the `stopped → starting` DB claim through recovery is correct, but the reclaim also held the in-process Phase-1 lock, which this module explicitly designs around never waiting under. Every elevated-tier restart also paid the uncached entitlement round-trip inside that lock, even when entitled.

Fixed in 8a42c5fb, with one deliberate deviation from the suggested fix: **skipping the reclaim when `lazy_init=True` alone would have silently removed enforcement** — every production restart is lazy, and the reaper only reconciles running rows, so a lapsed tier would never be reclaimed. Instead the reclaim moved to the unlocked Phase-2 continuation (your second suggestion):

- `_restart_workspace` now only marks elevated-tier restarts (`_pending_tier_recheck`, set where the row is already in hand — standard-tier restarts skip the check with zero extra reads).
- `_maybe_reclaim_lazy_tier` runs at the top of the Phase-2 owner block: the entitlement RTT overlaps the sandbox's background start, and on a lapse the destroy + recovery run with the lock released. Cross-worker arrivals wait on the existing `'starting'` claim machinery; same-worker callers coalesce on the Phase-2 event. A failed recovery re-arms `_pending_lazy_sync` so the generic Phase-2 handler reverts the row to `'stopped'` and surfaces the error.
- A wiring test asserts lock freedom directly (probes `lock.locked()` inside the recovery seam), plus live E2E: pro → performance workspace → stop → downgrade → `/start` reclaimed to standard with `SESSION_TIMING … lock_and_init=12ms tier_reclaim=4120ms` — the lock window is now 12ms where the old code held it for the full 4.1s.

The non-blocking timeout item is also in (1458118c): 120s on spec/always-on/duplicate/rename, 15s on the quota read — generous by design since a spec change legitimately runs ~15s and duplicates longer; the bound converts a hang into a visible failure rather than racing the server.

Chen-zexi:
Both items taken.

1. **Duplicate re-entry guard** — added (`handleDuplicateConfirm` now no-ops while busy). One calibration note: the confirm button was already `disabled={busy}`, and React 18 flushes discrete-event state updates before the next click is processed, so the two-copies scenario needed a window React effectively closes. The guard is in anyway — it matches `useWorkspaceMutation`'s dedupe semantics and removes the reliance on render timing.

2. **Router-level ownership on duplicate** — added `require_workspace_owner` like the nine sibling endpoints. This also unifies a semantic wrinkle the pattern-inconsistency hid: a non-owner previously got 404 (service `ValueError` mapping) where every sibling returns 403; now it's 403 across the board, with the service check kept as defense in depth. Added the endpoint's first router tests (403 non-owner, 201 happy path).

PR #307, opened 2026-07-05
fix(dispatch): fail loud when INTERNAL_SERVICE_TOKEN is unset instead of burning credits

dhishan:
## Problem

Running LangAlpha with an OAuth/subscription token, PTC dispatches consumed model credits but returned no output: the chat reported `dispatch_failed` while a full research workflow ran to completion in the background (observed: a single dispatch ran 34 minutes across 70 LLM calls). On OAuth/subscription tokens the spend is not tracked in the internal usage meter, so it stays invisible until the provider bill arrives.

## Root cause

Internal background dispatch (`X-Dispatch: background`) to `POST /api/v1/threads/{id}/messages` is only honoured when the request carries a matching `INTERNAL_SERVICE_TOKEN` (`is_internal` in `src/server/app/threads.py`). The dispatching caller (`ptc_agent` in `src/tools/secretary/tools.py`, and `_post_report_back` in `src/server/handlers/chat/report_back.py`) expects a JSON `{"status": "dispatched"}` ack and returns immediately.

When `INTERNAL_SERVICE_TOKEN` is unset the endpoint does not fail: it silently downgrades to the normal **foreground** path and returns a `text/event-stream` (SSE) `StreamingResponse`. That has two effects:

1. The full PTC/flash workflow runs synchronously in that request (minutes, many LLM calls), burning credits.
2. The caller cannot parse the SSE body as the JSON ack, so it reports `dispatch_failed` and the user sees nothing.

`INTERNAL_SERVICE_TOKEN` was also undocumented (absent from `.env.example`), so the default state of a fresh local deployment is exactly this silent credit sink.

## Fix

A dispatch that can never be honoured as background is a deployment misconfiguration, so fail loud at the source instead of firing a doomed, credit-burning request:

- `ptc_agent` aborts with `internal_service_token_missing` (before HITL / workspace creation) when the token is unset.
- `_post_report_back` drops (no retry, no HTTP) when the token is unset.
- Document `INTERNAL_SERVICE_TOKEN` in `.env.example`, noting it is required for background dispatch.

This is intentionally client-side. The endpoint still gracefully ignores a stray `X-Dispatch` header from a non-internal caller (existing `test_threads_authz.py` behaviour is unchanged); the guards only stop the internal callers from firing a request they know cannot be honoured.

## Tests

- New `test_dispatch_service_token_guard.py` pins the unset/blank-token abort for both call sites (no HTTP, no workspace side effects).
- Existing dispatch tests (`test_dispatch_workspace_cleanup.py`, `test_ptc_agent_continuation.py`, `test_post_report_back.py`) now set a token via a module autouse fixture, since a configured token is the normal production state and passing the new precondition is required to exercise the dispatch path.
- Full affected set: 60 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards so background dispatch now stops early when required internal authentication is missing, preventing unintended foreground runs and wasted processing.
  * Improved error handling to return a clear failure instead of attempting a dispatch that can’t be authenticated.

* **Documentation**
  * Updated the environment setup example to include the new required configuration and guidance.

* **Tests**
  * Expanded test coverage for missing/blank authentication and updated existing tests to run with the required test configuration.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -127,6 +127,13 @@ AUTH_USER_ID=local-dev-user
 # Set to your public-facing URL in production (e.g. https://app.example.com)
 # SERVER_BASE_URL=http://localhost:8000
 
+# Shared secret authenticating internal service-to-service calls (the backend
+# dispatching PTC/report-back workflows to its own /messages endpoint). REQUIRED
+# for background dispatch: when unset, dispatch cannot be authenticated and the
+# workflow runs in the foreground, burning model credits with no result.
+# Generate manually: openssl rand -hex 32
+INTERNAL_SERVICE_TOKEN=
+
 # =============================================================================
 # Infrastructure — controls whether Docker Compose starts PostgreSQL and Redis
 # =============================================================================
--- a/src/server/handlers/chat/report_back.py
+++ b/src/server/handlers/chat/report_back.py
@@ -958,6 +958,21 @@ async def _post_report_back(
 
     self_base_url = os.environ.get("GINLIXFLOW_BASE_URL", "http://localhost:8000")
     service_token = os.environ.get("INTERNAL_SERVICE_TOKEN", "")
+    # Same invariant as the ptc_agent dispatch tool: report-back POSTs to the
+    # internal /messages endpoint with ``X-Dispatch: background``, honoured only
+    # when ``INTERNAL_SERVICE_TOKEN`` matches. With the token unset the endpoint
+    # runs the flash summary in the *foreground* and streams SSE, burning
+    # credits for a report-back the caller can never consume. Drop loudly rather
+    # than dispatch a doomed, credit-burning request.
+    if not service_token.strip():
+        logger.error(
+            "[FLASH_REPORT_BACK] INTERNAL_SERVICE_TOKEN is not set; cannot "
+            "dispatch report-back for PTC thread %s without burning credits on "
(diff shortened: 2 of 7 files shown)


Chen-zexi:
Thanks for pointing this out.

One note on the incident itself: the run was not wasted. When the dispatch fell back to a foreground run, the PTC agent still executed to completion, and its output is viewable in the workspace it dispatched into. What broke was the handoff: the flash side reported dispatch_failed and never got the result back. That part is indeed a bug.

On the fix, I went a different way than requiring the token on the client. In oss mode the token should not gate dispatch at all (auth is disabled there), so requiring it would break the zero config setup. The fix lands at the endpoint instead: with no token configured in oss mode the self dispatch is trusted, and an unauthenticated background dispatch is rejected with 403 rather than silently run in the foreground. A token set in oss mode is still enforced. Your preflight guards are folded in, conditioned to auth enabled mode, so a doomed dispatch fails loud before any side effect.

The fix is up in #308 with you as co-author for spotting the issue and proposing the fix. Closing this PR in favor of that change.

dhishan:
Thanks @Chen-zexi - your endpoint-level approach in #308 is the right call (keeping OSS zero-config working while still rejecting unauthenticated background dispatch). Closing in favor of #308.

PR #322, opened 2026-07-13
fix: recover from sandbox transport failures and 404 malformed thread ids

Chen-zexi:
## Summary

Two production hardening fixes on the request/replay path:

1. **Sandbox transport errors are now recoverable.** When Daytona's transport
   flakes mid-turn, the sandbox layer raises `SandboxTransientError` /
   `SandboxGoneError`. `classify_error` matched on error-message substrings and
   missed these types, so a transient transport blip was classified as
   non-recoverable and killed the turn with a terminal `[PTC_ERROR]` instead of
   handing off to the existing workflow auto-retry. It now classifies these
   exceptions by type as recoverable, so the turn lands `interrupted`
   (resumable) and the client-driven retry path can pick it up.

2. **Malformed thread ids return 404, not 500.** `get_thread_by_id` and
   `get_replay_thread_data` bound the raw `thread_id` straight into the query,
   so a non-canonical or non-UUID id (e.g. a filename that reached a thread
   route) raised Postgres `22P02` and surfaced as an HTTP 500. Both paths now
   run the id through `normalize_uuid` first and return "not found" for
   anything that isn't a UUID — closing the two binds missed by the earlier
   normalization sweep. The web client mirrors this: workspace ids from the URL
   and navigation state are sanitized once at the route boundary in
   `ChatAgent`, so a malformed id never reaches a workspace route or API call.

## Overview

**Net change (excluding tests):** 9 files, +73 / −14.

| | Files | +/− |
|---|---|---|
| Product code | 8 | +60 / −14 |
| Tests + e2e fixtures | 9 | +251 / −89 |
| **Total** | **17** | **+311 / −103** |

Most of the test/fixture volume is a pure e2e id swap (`chat.spec.js` 60/60,
`mockResponses.js` 19/19 — placeholder ids → valid UUIDs), which nets to ~0.

**By module:**

- **Backend — error classification** (`src/server/handlers/chat/_common.py`, +11/−2):
  classify `SandboxTransientError`/`SandboxGoneError` as recoverable.
- **Backend — thread id normalization** (`src/server/database/conversation.py`, +8;
  `src/tools/secretary/tools.py`, +3/−4): `normalize_uuid` on the two
  remaining raw binds; freshen a now-stale comment.
- **Backend — agent.md fault tolerance** (`src/ptc_agent/core/session.py`, +7):
  log read failures in the `except Exception` handler (missing files already
  return `None` from `aread_file_text` and never raise), and record that
  invariant in a comment.
- **Web — workspace-id boundary** (`ChatAgent.tsx` +14/−7, `ChatView.tsx` +5,
  `ThreadGallery.tsx` +2/−1, `utils/uuid.ts` +10 new): sanitize once at the
  route boundary via Zod `safeParse`; graceful `/chat` fallback + console
  warning instead of silent drops.

**What this introduces:** no new endpoints, config, or dependencies. One new
web util (`isValidUuid`, Zod-backed) and one behavioral change to error
classification; everything else tightens existing binds and removes dead code.

## Diff Size

17 files, +311 / −103. Product code (excl. tests + e2e fixtures): 8 files,
+60 / −14. The remaining 9 files are unit tests + e2e fixtures, dominated by
the placeholder-id → UUID swap that nets to ~0.

## Test Coverage

- Backend unit suite: **6058 passed, 1 xfailed** (30s). Ruff clean on all
  changed files.
- Web: **2394 passed** + **42 e2e**. `tsc --noEmit` clean; ESLint 0 errors
  in changed files.
- New/updated tests:
  - `test_session_agent_md.py` (new): missing-file → `None`, transient-error →
    `None` + recovery after invalidate, cache-until-invalidated.
  - `test_pg_sanitize.py`: `normalize_uuid` canonical / non-canonical
    (uppercase, `urn:uuid:`, hyphenless, braced) / non-UUID contract.
  - `test_chat_common.py`, `test_conversation_db.py`,
    `test_threads_malformed_id.py`: sandbox-error classification + malformed-id
    404 on replay/fetch.
  - `uuid.test.ts` (new): `isValidUuid` guard (Zod `safeParse`).
- **Red/green reproduction:** reverting `_common.py` + `conversation.py` to
  base flips the 5 replay/classification regression tests to failing
  (`assert 500 == 404`); restoring turns them green.

## Pre-Landing Review

Multi-specialist + adversarial pass. Findings: design 5, testing 4, security 2,
maintainability 0, performance 0 — all actionable ones applied (route-boundary
sanitization consolidation, dead-branch removal, added coverage).

**Known pre-existing gaps in the shared retry rails** (documented, not
introduced here; timeout/network/api errors already ride these same rails in
prod today — deferred to the turn-lifecycle rework):

- The web client has no handler for the `retry` SSE event / `auto_retry` flag,
  so recovery is human-in-the-loop (the thread lands `interrupted` and the user
  resumes) rather than automatic.
- `increment_retry_count` returns 0 when no tracker status exists yet, so a
  failure during turn setup (before the turn is marked active) doesn't count
  against the retry cap.
- A recoverable setup-phase failure can leave an orphan in-progress query row.

The net user-facing win of this PR stands independent of those: a transient
sandbox blip now yields a resumable `interrupted` thread instead of a dead turn.

## Plan Completion

6 done / 4 deferred / 0 not-done. Deferred items are the retry-rail gaps above
plus the polish items scoped out at the start (jitter, config-lift, message
copy).

## Test plan

- [x] `uv run pytest tests/unit/ -q` — 6058 passed, 1 xfailed
- [x] `cd web && pnpm test` — 2394 passed
- [x] `cd web && pnpm test:e2e` — 42 passed
- [x] `cd web && pnpm typecheck` — clean
- [x] Ruff clean on changed files
- [x] Red/green regression reproduction verified

### DIFF
--- a/src/ptc_agent/core/session.py
+++ b/src/ptc_agent/core/session.py
@@ -71,6 +71,13 @@ class Session:
                         self.sandbox.normalize_path("agent.md")
                     )
                 except Exception:
+                    # A missing agent.md is a silent None from aread_file_text;
+                    # reaching here means the read itself failed.
+                    logger.warning(
+                        "Failed to read agent.md",
+                        conversation_id=self.conversation_id,
+                        exc_info=True,
+                    )
                     self._agent_md_cache = None
             else:
                 self._agent_md_cache = None
--- a/src/server/database/conversation.py
+++ b/src/server/database/conversation.py
@@ -2007,6 +2007,10 @@ async def get_thread_by_id(conversation_thread_id: str) -> Optional[Dict[str, An
     Returns:
         Thread dict or None if not found
     """
+    conversation_thread_id = normalize_uuid(conversation_thread_id)
+    if conversation_thread_id is None:
+        return None
+
     try:
         async with get_db_connection() as conn:
             async with conn.cursor(row_factory=dict_row) as cur:
@@ -2944,6 +2948,10 @@ async def get_replay_thread_data(
     Returns:
         (owner_user_id, thread_summary, queries, responses, usages, provenance)
     """
+    thread_id = normalize_uuid(thread_id)
+    if thread_id is None:
+        return None, None, [], [], [], []
+
     async with get_db_connection() as conn:
         async with conn.cursor(row_factory=dict_row) as cur:
             # 1. Owner check (join threads -> workspaces)
--- a/src/server/handlers/chat/_common.py
+++ b/src/server/handlers/chat/_common.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Any, Optional
 import psycopg
 from fastapi import HTTPException
 
+from ptc_agent.core.sandbox.runtime import SandboxGoneError, SandboxTransientError
 from src.config.settings import (
     get_langsmith_metadata,
     get_langsmith_tags,
@@ -256,6 +257,8 @@ def classify_error(e: Exception) -> dict:
     is_non_recoverable = isinstance(e, non_recoverable_types)
 
     # Recoverable error patterns (transient issues)
+    is_sandbox_error = isinstance(e, (SandboxTransientError, SandboxGoneError))
+
     is_postgres_connection = isinstance(
         e, psycopg.OperationalError
     ) and "server closed the connection" in str(e)
@@ -300,12 +303,18 @@ def classify_error(e: Exception) -> dict:
     )
 
     is_recoverable = (
-        is_postgres_connection or is_timeout or is_network_issue or is_api_error
+        is_sandbox_error
+        or is_postgres_connection
+        or is_timeout
+        or is_network_issue
+        or is_api_error
     ) and not is_non_recoverable
 
     # Determine specific error_type label
     if is_recoverable:
-        if is_postgres_connection or is_network_issue:
+        if is_sandbox_error:
+            error_type = "transient_error"
+        elif is_postgres_connection or is_network_issue:
             error_type = "connection_error"
         elif is_timeout:
             error_type = "timeout_error"
--- a/src/tools/secretary/tools.py
+++ b/src/tools/secretary/tools.py
@@ -427,10 +427,9 @@ async def ptc_agent(
         from src.server.database.conversation import get_thread_by_id
         from src.server.utils.pg_sanitize import normalize_uuid
 
-        # Normalize once so the owner check and the fetch bind the same canonical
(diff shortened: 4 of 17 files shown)


Chen-zexi:
Both findings addressed:

**#1 (session.py) — corrected.** Right that the net diff vs `main` shows only the `logger.warning` addition. The `except FileNotFoundError` branch was added and removed within this branch, so it nets out — the "drops an unreachable branch" wording described the intra-branch action, not the net change. Fixed the PR body's Overview to describe the net change accurately: log read failures in the `except Exception` handler (missing files already return `None` from `aread_file_text`), plus a comment recording that invariant.

**#2 (ChatView stale panel) — intentional no-op.** When an artifact ref carries an invalid workspace id, ignoring it (leaving any open panel as-is) is deliberate: tearing down a legitimately-open panel because one malformed ref arrived is worse than dropping the ref. Now logs a `console.warn` so the drop is visible rather than silent.

Also fixed the e2e chat suite (4 failures): the fixtures used non-UUID ids (`ws-1`/`th-1`) that the new route-boundary guard correctly rejects, so they no longer navigated. Replaced with valid UUIDs — suite is green (42 passed).

Chen-zexi:
Addressed the two majors; keeping the nitpick as-is.

**Fixture corruption (Major) — fixed.** The blanket id replace was substring-greedy: it rewrote `th-2` inside "NVIDIA Earth-2" and `ws-1` inside `news-1`. Reverted the four fixture files and redid the swap with word boundaries (`perl -pi -e 's/\bws-1\b/…/g'`), so only standalone ids change. Verified `news-1` and the "Earth-2" product name are intact, no UUID fragment is glued to surrounding word characters, and the replacement counts match the word-boundary token counts (18 `ws-1`, 61 `th-1`). E2E green.

**Hand-rolled UUID regex (Major) — fixed.** Switched `isValidUuid` to `z.string().uuid().safeParse()`, matching the boundary-validation convention (Zod `safeParse`, never throws, never `.parse()` at a boundary). Real ids are v4 so Zod's stricter version/variant check doesn't reject anything valid; existing unit tests still pass.

**All-or-nothing abort (nitpick) — intentional, keeping.** When an artifact ref names a *different* workspace whose id is malformed, falling back to the current workspace would risk opening the file in the wrong workspace. Dropping the ref (now with a `console.warn`) is the safer default. Left as-is.

PR #331, opened 2026-07-24
Add Atlas Cloud LLM provider

binyangzhu000-sudo:
## Summary
- Register Atlas Cloud as an OpenAI-compatible LLM provider using https://api.atlascloud.ai/v1.
- Add visible Atlas Cloud model entries for qwen/qwen3.5-flash and deepseek-ai/deepseek-v4-pro.
- Support ATLAS_CLOUD_API_KEY as a fallback alias for ATLASCLOUD_API_KEY while keeping explicit BYOK overrides first.
- Cover provider registration, model manifest entries, and env alias precedence with focused unit tests.

## Validation
- uv run pytest tests/unit/llms/test_system_provider_and_tiers.py tests/unit/llms/test_model_config_flatten.py tests/unit/llms/test_manifest_integrity.py (259 passed)
- uv run python -m compileall src/llms tests/unit/llms
- uv run --extra dev ruff check src/llms/llm.py tests/unit/llms/test_system_provider_and_tiers.py
- python3 -m json.tool src/llms/manifest/models.json
- python3 -m json.tool src/llms/manifest/providers.json
- git diff --check and git diff --cached --check
- Atlas Cloud live model catalog: 437 total / 375 visible, confirmed qwen/qwen3.5-flash and deepseek-ai/deepseek-v4-pro

## README / promotion
No README, docs, logo, sponsor, credits, or partner copy changes.

## Summary by CodeRabbit

* **New Features**
  * Added Atlas Cloud as a supported AI provider.
  * Added Atlas Cloud models: Qwen 3.5 Flash and DeepSeek V4 Pro.
  * Added support for alternative API key environment variable names.
* **Bug Fixes**
  * Improved API key resolution when the primary environment variable is unset.
  * Updated error messages to list all expected API key environment variables.
* **Tests**
  * Added coverage for API key aliases and Atlas Cloud model configuration.


binyangzhu000-sudo:
Synced with `main` — this branch was 418 commits behind.

The only conflict was `src/llms/manifest/models.json`, where this branch's two `atlascloud` entries landed in the same spot as new upstream models. I resolved it at the JSON level rather than as text: took your file as the base and re-added only the two keys this branch introduces. That matters here because upstream retired **38** model keys in the meantime — a textual resolution could easily have resurrected some of them.

Verified the merged manifest rather than assuming:
- the `atlascloud` model ids in `models.json` and `providers.json` match each other exactly
- all **54** upstream models are present and byte-identical (0 missing, 0 modified)
- upstream's `provider_config` and `models` groupings are unchanged — `atlascloud` is purely additive

I couldn't run the unit suite here: `tests/unit/llms/` pulls in `langchain_openai` and the rest of the runtime stack, which my environment doesn't have. Since the change is manifest data, the checks above target exactly what it touches — but CI is the authority.

Disclosure: I work at Atlas Cloud.

PR #337, opened 2026-07-30
fix: canonicalize the sandbox state so the settings panel stops hiding live tabs

### LINKED ISSUE #333: packages and tools & skills in Sandbox Settings don't show up even sandbox works normal
The frontend’s intended logic is:
```
type SandboxState =
  | 'running'
  | 'stopped'
  | 'archived'
  | 'starting'
  | 'unknown';
```

`const isRunning = stats?.state === 'running';`
When Workspace Settings opens:
Request [/api/v1/workspaces/{workspaceId}/sandbox/stats](url)
Read the returned state.
If it is running:Show “Running” and a Stop button.
Display Storage and Packages.
Display connected MCP servers and available skills.

If it is stopped or archived:Show only offline metadata.
Tell the user to start or restore the workspace.

Refresh the stats after start, stop, restore, or sync operations.
The correct flow is:
```
Backend returns state: "running"
        ↓
isRunning = true
        ↓
Display MCP servers and available skills
        ↓
yingmi-skill becomes visible
```
The bug existed because the frontend checked for "started" instead of "running".

Chen-zexi:
## Overview

A **running sandbox reported itself as not-running**, so the settings panel hid the Packages, Tools & Skills and MCP tabs. Fixes #333. The gate fix is three lines; the rest is the fallout of finding out *who* the bug was actually hitting. Carries one unrelated commit — an `mcp` pin that unblocks the sandbox image build (see below).

| | Files | Added | Removed |
|---|---|---|---|
| Modified (production) | 5 | 234 | 48 |
| New (production) | 0 | — | — |
| **Net excl. tests** | **5** | **+234** | **−48** |
| Tests (1 new, 2 modified) | 3 | 668 | 1 |
| **Total** | **8** | **+902** | **−49** |

**By module**

| Module | File | Δ | What |
|---|---|---|---|
| `src/server/app/` | `workspace_sandbox.py` | +75/−5 | `_display_state()` adapter at the API boundary; offline path never claims `running`; `provider` on the wire, read from config |
| `web/…/components/` | `SandboxSettingsPanel.tsx` | +130/−36 | Gate on canonical `'running'`; terminal-state allowlist + label map; always-enabled Refresh + newest-response-wins guard; provider-aware Storage; auto-stop and a11y fixes |
| `src/ptc_agent/core/sandbox/providers/` | `docker.py` | +24/−5 | `get_metadata` returns cpu/memory/created_at from the inspect call it already made; status parse shared with `get_state` |
| `Dockerfile.sandbox`, `core/sandbox/` | `_defaults.py` | +5/−2 | Pin `mcp==1.29.0` in both mirrored dependency lists |
| `tests/unit/server/app/` | `test_workspace_sandbox_stats.py` | +395 (new) | 24 tests — canonicalization, the offline clamp, both fallback branches, provider sourcing, ownership |
| `web/…/__tests__/` | `SandboxSettingsPanel.test.tsx` | +241/−1 | 20 tests — gated tabs, fail-safe states, label leaks, auto-stop, provider-aware Storage, the synonym boundary, refresh, response ordering |
| `tests/unit/core/sandbox/` | `test_docker_provider.py` | +32 | Docker metadata limits, and that unset limits are omitted rather than reported as 0 |

**What this introduces:** two adapter functions, one wire field (`provider`), a state→label map, one button, and a request-sequence guard. No new endpoint, dependency, migration, or config.

## The bug

`SandboxRuntime.get_metadata()` is provider-specific by contract (`runtime.py:271-272`). For the same healthy sandbox:

- **Daytona** (hosted) reports `"started"`
- **Docker** (local self-host) reports `"running"`

The endpoint forwarded that raw value and the panel compared `state === 'started'`. The gate was reading a provider dialect as if it were a shared vocabulary.

## Who it was actually hitting

Not symmetrically — and not the way the issue reads:

- **Self-hosted (Docker): always broken.** `DockerRuntime.get_metadata` derives state from `get_state()`, which can only return `RuntimeState` values — never `"started"`. `isRunning` was therefore *structurally, permanently false*. Storage, Packages and Tools & Skills have never rendered once outside hosted.
- **Hosted (Daytona): broken only when the metadata read failed.** The raw `"started"` matched the old gate, so the ordinary path worked. But the full-stats path seeded `state = None` and swallowed metadata errors with a bare `except: pass` — so any read failure silently downgraded a live sandbox to "offline", with no log line to explain it.

So the gate fix and the fallback fix are doing two different jobs: one makes three tabs reachable for self-hosted users for the first time, the other removes a silent hosted failure mode. That is why the PR does not stop at the gate.

## The fix

**1. Canonicalize at the API boundary** (not in the provider):

```python
_DISPLAY_STATE_SYNONYMS = {"started": "running"}
```

The boundary is the right owner precisely *because* `get_metadata()` is documented provider-specific — normalizing inside `DaytonaRuntime` would make one provider secretly speak another's dialect and leave the next provider to rediscover the trap. Only the one synonym is rewritten; everything else passes through, so this layer never enumerates Daytona's ~23 states.

**The offline path is the exception: it never reports `running` at all.** That path answers for a workspace whose row says `starting`/`stopping`, and it collects no disk usage, packages or skills — while the client reads `running` as proof they are present and as licence to offer Stop, which `stop_workspace` then rejects for any non-running row. So when the provider says the sandbox is up but the row disagrees, the row wins; every other provider state still passes through. The full path takes over the moment the row says running.

**2. Invert the frontend state list.** `TRANSITIONAL_STATES` (3 entries) became `TERMINAL_STATES` (9 settled values) plus a `STATE_LABELS` map. Previously any state outside those three — `creating`, `restoring`, `resizing`, `destroying`, `pulling_snapshot`, … — rendered as a settled failure with a **working Start button**, so a user could start a mid-resize sandbox. Unrecognized states now fail safe to in-progress with actions disabled. This also stops two identifiers reaching the screen as copy: `build_failed` → "Build_failed", and Daytona's `_missing_` sentinel → "Unknown_default_open_api".

Failing safe has one cost that has to be paid explicitly: **nothing polls** — stats load on mount and after explicit actions only. So the status row gains an always-enabled **Refresh**. In a transitional state every other control is disabled, and without it the panel has no way to advance short of a remount.

That affordance brings its own hazard, so it is paid for in the same change: `handleStartStop` never sets `loading`, so Refresh stays clickable while a Stop is in flight. The read it starts takes the full path and its ~15s of probes while the post-stop read takes the fast offline one, and committing by arrival order let the stale response resurrect a stopped sandbox as running. A monotonic request id now lets only the newest response commit.

The client deliberately does **not** carry its own `'started'` shim. Canonicalization belongs to the API alone — two layers each deciding what "running" means is how the vocabularies drift apart, and the client would have no way to know which one was authoritative. A test pins that absence so it doesn't get re-added casually.

**3. Make the newly-reachable tabs honest for Docker.** `get_metadata` now returns cpu/memory/created_at (from the `container.show()` call it already made), so the Overview cards aren't four `---`. And `provider` is on the wire because the Storage tab must suppress its totals for Docker: `df -h` inside a container with no size quota reports the **host** filesystem, so a self-hosted user would see their laptop's disk pressure labelled as their sandbox's. The per-directory breakdown is sandbox-scoped and stays.

`provider` is read from the configured sandbox provider, *not* from `meta["provider"]` — only the Docker provider sets that key, and the metadata read can raise. Either would report "not docker" on the exact path the field exists to protect.

**4. Three adjacent corrections.** `auto_stop_interval == 0` renders "Always on" — 0 *disables* auto-stop, so "Auto-stop: 0m" told the user the exact opposite of what the sandbox would do. `role="status" aria-live="polite"` on the status row, `aria-hidden` on the decorative dot and spinner. `focus-visible` outlines on 7 inputs that had bare `outline-none` and no replacement, including every Vault credential field.

## The unrelated commit: pinning `mcp`

`ff107fea` is not part of the fix. `mcp 2.0.0` (2026-07-28) dropped the `mcp.server.fastmcp` module that scrapling's MCP entrypoint imports, and the dependency was unpinned in both mirrored lists — so every sandbox image build since has failed. The 1.29.0 wheel ships 19 files under `mcp/server/fastmcp/`; the 2.0.0 wheel ships zero. CI history shows no run of the sandbox workflow between the release and this branch, which is why nothing caught it earlier; `main` fails the same way today.

Pinned to `mcp==1.29.0` rather than `mcp<2` because `mcp_setup` joins `DEFAULT_DEPENDENCIES` into a shell command, where `<` parses as a redirect and would quietly install `mcp` unpinned again. Verified by resolution, not assumption: `mcp==1.29.0`, `fastmcp==3.4.5` and `scrapling==0.4.12` resolve together. This is a stopgap — the 2.x migration is a separate PR.

## Verification

- Backend `tests/unit/` (7152 passed) and frontend `vitest run` (2534 passed) both fully green; `tsc --noEmit` and `ruff check` clean.
- **Every new behaviour was mutation-tested, not just covered.** Each production change was reverted in turn and the intended test failed, with checksum-verified restores:
  - offline path allowed to report `running` again → both clamp tests fail
  - drop the offline metadata fallback → caught by the no-state-key test *(this branch previously survived deletion with the whole suite green)*
  - offline exception fallback → `None` → caught by the provider-failure test
  - treat the 4 in-progress states as terminal → the 4 fail-safe tests fail
  - restore the bare capitalize → the identifier-leak test fails
  - always render the raw auto-stop interval → the "Always on" test fails
  - show Docker the host disk totals → the Storage suppression test fails
  - drop `meta["cpu"]` → the Docker limits test fails
  - re-add a client-side `'started'` shim → both synonym-boundary tests fail, so the absence is pinned rather than merely current
  - disable Refresh while transitioning → the refresh test fails
  - remove the newest-response-wins guard → the response-ordering test fails
  - source `provider` from metadata again → 3 provider tests fail, one of them by crashing on the unbound `meta` in the exception path
- **End-to-end on a running stack against live Daytona**, not just unit-mocked. Booted the backend, provisioned a real sandbox, and hit the endpoint over HTTP:

  ```
  GET /api/v1/workspaces/{id}/sandbox/stats
  → "state": "running", "provider": "daytona", "auto_stop_interval": 60,
    resources {cpu 1.0, memory 1.0, disk 3.0}, disk_usage {total "3.0G", used "6.3M"},
    directory_breakdown [...6 entries...]
  ```

  Then read the raw provider value for that same sandbox to prove the adapter did the work:

  ```
  RAW provider state          'started'      ← what Daytona actually reports
  after _display_state        'running'      ← what reaches the wire
  meta has a "provider" key    False         ← why provider is read from config
  configured provider         'daytona'
  ```

  The `meta` result is worth calling out: Daytona genuinely never sets a `provider` key, so sourcing that field from metadata would have silently reported "not docker" — confirmed by observation rather than by reading the provider source.
- **Docker's `get_metadata` exercised against real containers** — the shipped code against a throwaway container created with real cgroup limits (`NanoCpus=2e9`, `Memory=4GiB`): returned `cpu: 2.0`, `memory: 4.0`, a real `created_at`, and no `disk` key. A second container with no limits omitted `cpu`/`memory` rather than reporting `0`. It also reported `state: "running"`, confirming by observation that Docker can never emit `"started"`. The full self-hosted stack was not exercised end-to-end — `sandbox-docker` in CI covers that.
- Ownership pinned on both paths; the docstring records that only the `stopped` case pins the endpoint-level check, since the running path re-checks inside `_get_sandbox`.
- Each of the 4 commits was verified green independently (backend unit at every slice), and the consolidation was verified as a pure reshaping — identical tree, empty diff against the pre-consolidation branch.

## Scope check

- **This endpoint has no consumer outside this service.** Swept every sibling codebase we deploy alongside langalpha: zero references to `sandbox/stats`, zero sandbox-state `"started"` comparisons. The settings panel is the only caller.
- No JSON fixture or snapshot holds a sandbox state of `"started"`. Raw-SDK fixtures exercise `_STATE_MAP`/the CLI, not this endpoint, and are untouched.
- Hosted risk is limited by construction — the only `src/ptc_agent/` changes are `docker.py`, the local-only provider, and the pinned dependency list — and a live hosted sandbox still reports as running through the real endpoint.

## Summary by CodeRabbit

- **New Features**
  - Added sandbox provider information to workspace statistics.
  - Added a manual refresh control for sandbox status.
  - Added clearer status labels and progress indicators, including “Always on” auto-stop settings.
  - Added provider-specific storage usage messaging for Docker sandboxes.
  - Added resource and creation-time details where available.

- **Bug Fixes**
  - Improved status consistency across sandbox providers.
  - Prevented outdated status responses from overwriting newer information.
  - Improved fallback behavior when sandbox details are unavailable.
  - Enhanced accessibility for live status updates and loading indicators.

### DIFF
--- a/Dockerfile.sandbox
+++ b/Dockerfile.sandbox
@@ -92,7 +92,7 @@ ENV PLAYWRIGHT_BROWSERS_PATH=/usr/local/ms-playwright
 RUN echo 'curl_cffi>=0.14' > /tmp/overrides.txt \
     && uv pip install --system \
     --override /tmp/overrides.txt \
-    mcp fastmcp fastapi \
+    mcp==1.29.0 fastmcp fastapi \
     pandas requests aiohttp "httpx[http2]" \
     numpy scipy scikit-learn statsmodels \
     yfinance \
--- a/src/ptc_agent/core/sandbox/_defaults.py
+++ b/src/ptc_agent/core/sandbox/_defaults.py
@@ -10,7 +10,10 @@ SANDBOX_NODE_VERSION = "24.14.1"  # Pinned; mirrored in Dockerfile.sandbox.
 
 DEFAULT_DEPENDENCIES = [
     # Core
-    "mcp",
+    # Pinned to 1.x: mcp 2.0.0 dropped ``mcp.server.fastmcp``, which scrapling's
+    # MCP entrypoint imports. Exact pin rather than "mcp<2" because mcp_setup
+    # joins this list into a shell command — "<" would parse as a redirect.
+    "mcp==1.29.0",
     "fastmcp",
     "fastapi",
     "pandas",
--- a/src/ptc_agent/core/sandbox/providers/docker.py
+++ b/src/ptc_agent/core/sandbox/providers/docker.py
@@ -150,10 +150,14 @@ class DockerRuntime(SandboxRuntime):
             pass
         await self._container.delete(force=True)
 
-    async def get_state(self) -> RuntimeState:
-        info = await self._container.show()
+    @staticmethod
+    def _state_from_inspect(info: dict[str, Any]) -> RuntimeState:
+        """Shared by get_state and get_metadata so the unknown-status fallback stays one rule."""
         status = info.get("State", {}).get("Status", "unknown")
         return _DOCKER_STATE_MAP.get(status, RuntimeState.ERROR)
+
+    async def get_state(self) -> RuntimeState:
+        return self._state_from_inspect(await self._container.show())
 
     # -- Execution --
 
@@ -568,14 +572,29 @@ class DockerRuntime(SandboxRuntime):
         raise NotImplementedError("Docker provider does not support archive")
 
     async def get_metadata(self) -> dict[str, Any]:
-        state = await self.get_state()
-        return {
+        """One inspect call serves state, limits and creation time.
+
+        Limits read back as 0 on a container created outside this provider's
+        config, so report nothing rather than a bogus zero. No ``disk`` key —
+        docker sets no size quota, so there is no sandbox-scoped figure to give.
+        """
+        info = await self._container.show()
+        host_config = info.get("HostConfig") or {}
+
+        meta: dict[str, Any] = {
             "id": self._id,
             "working_dir": self._working_dir,
-            "state": state.value,
+            "state": self._state_from_inspect(info).value,
             "dev_mode": self._dev_mode,
             "provider": "docker",
         }
+        if created := info.get("Created"):
+            meta["created_at"] = created
+        if nano_cpus := host_config.get("NanoCpus"):
+            meta["cpu"] = round(nano_cpus / 1e9, 2)
+        if memory := host_config.get("Memory"):
+            meta["memory"] = round(memory / 1024**3, 2)
+        return meta
 
     # -- Internal: tar-based file I/O --
 
--- a/src/server/app/workspace_sandbox.py
+++ b/src/server/app/workspace_sandbox.py
@@ -39,6 +39,60 @@ router = APIRouter(prefix="/api/v1/workspaces", tags=["Workspace Sandbox"])
 
 _SIGNED_URL_TTL = 3000  # 50 min (signed URLs expire in 1h)
 
+# Provider-native synonyms for "up and usable", canonicalized for the wire.
+# get_metadata() is provider-specific by contract, so daytona reports "started"
+# where docker reports "running". Daytona's own _STATE_MAP maps the same synonym
+# one layer down for RuntimeState — renaming a state there needs a change here too.
+_DISPLAY_STATE_SYNONYMS = {"started": "running"}
+
+
+def _configured_provider() -> str | None:
+    """Which provider this deployment runs, read from config rather than metadata.
+
+    Deliberately not ``meta["provider"]``: only the docker provider sets that key,
+    and the metadata read can fail — either would silently report "not docker" and
+    defeat the disk-quota branch this value exists to drive.
+
+    Reads ``config.sandbox`` directly rather than via ``to_core_config()``, which
+    deep-copies the MCP config to hand each workspace its own — wasted work here,
+    since it passes ``sandbox`` straight through.
+    """
+    try:
+        config = WorkspaceManager.get_instance().config
+        provider = getattr(getattr(config, "sandbox", None), "provider", None)
+        return provider if isinstance(provider, str) else None
+    except Exception as e:
+        logger.debug(f"Could not resolve the configured sandbox provider: {e}")
+        return None
+
+
+def _display_state(state: Any) -> str | None:
+    """Canonicalize a provider state for the wire; see _DISPLAY_STATE_SYNONYMS.
+
+    Unwraps enums defensively: ``get_metadata`` is provider-specific by contract,
+    and ``str()`` on a str-mixin enum yields ``"RuntimeState.RUNNING"`` rather
+    than its value. Both current providers already stringify.
+    """
+    if not state:
+        return None
+    raw = str(getattr(state, "value", state))
+    return _DISPLAY_STATE_SYNONYMS.get(raw, raw)
+
+
+def _offline_display_state(state: Any, row_status: str | None) -> str | None:
+    """Like _display_state, but never reports "running" — the offline path can't back it.
+
+    That path collects no disk usage, packages or skills, while the client reads
+    "running" as proof they are present and as licence to offer Stop. When the
+    provider says the sandbox is up but the row still says starting/stopping, the
+    row is the honest answer: it is what the action endpoints validate against,
+    and the full path takes over the moment it says running.
+    """
+    display = _display_state(state)
+    if display is None or display == "running":
+        return row_status
+    return display
+
 
 def _preview_cache_key(sandbox_id: str, port: int) -> str:
     """Redis key for cached signed preview URL."""
@@ -163,9 +217,19 @@ class SkillInfo(BaseModel):
 
 
 class SandboxStatsResponse(BaseModel):
+    """Sandbox stats for the workspace settings panel.
+
+    ``state`` is a display vocabulary, not ``RuntimeState``: "running" is canonical
+    across providers, anything else is a provider or workspace value passed through
+    as a label.
+    """
+
     workspace_id: str
     sandbox_id: str | None = None
     state: str | None = None
+    # Which provider answered. The UI needs it because disk_usage is a df(1) read
+    # inside the sandbox, and docker sets no disk quota, so its totals are the host's.
+    provider: str | None = None
     created_at: str | None = None
     auto_stop_interval: int | None = None
     resources: SandboxResources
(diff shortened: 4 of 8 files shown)


Chen-zexi:
All six findings from the three reviews were verified against the code. Five are fixed as proposed; one is fixed differently, and half of one is refuted. Replying in one place because the branch was since consolidated 8 → 4 commits and force-pushed, which marks the inline anchors outdated.

| Finding | Verdict | Fix |
|---|---|---|
| Codex P2 — refresh races stale responses | Confirmed (one half refuted) | `5d8a9e9c` |
| Codex P2 — offline path advertises `running` | Confirmed; fixed via the second option | `71b75601` |
| claude — redundant `to_core_config()` | Confirmed | `71b75601` |
| claude — `provider` inconsistent across offline branches | Confirmed, provably harmless | `71b75601` |
| coderabbit — `provider` typed non-nullable | Confirmed (`provider` only) | `5d8a9e9c` |
| coderabbit — status parse duplicated with `get_state` | Confirmed | `b8f318b3` |

**Refresh race — the "presses it repeatedly" half is not reachable.** `loading` renders `<LoadingSkeleton/>`, which unmounts the Refresh button, so the second click has nothing to land on; a test written for that scenario fails on the call count rather than the race. The reachable half is the Stop interleave, and it is exactly as described: `handleStartStop` never sets `loading`, so Refresh survives on screen for the duration of the POST, its read takes the full path (~15s of probes per the `asyncio.gather` timeouts) while the post-stop read takes the fast offline path, and arrival order decided the winner.

Fixed with a monotonic request id rather than by disabling Refresh during `actionLoading`. Disabling it would weaken the affordance it exists to provide — in a transitional state every other control is already disabled and nothing polls — and it would not cover the general case of two overlapping reads. The guard makes only the newest response commit, whatever started it.

**Offline path — took the second suggestion, not the first.** Confirmed end-to-end: `stop_workspace` raises for any row that isn't `running`, so the enabled Stop really does fail. Rather than add a data-readiness field, the offline path now never reports `running` at all: it structurally cannot back that claim, so it defers to the workspace row, while every other provider state still passes through. That removes the inconsistency without adding a second field for the client to keep in sync.

Worth recording that this was not a regression in this PR. On `main` the offline path returned `state=meta.get("state")` raw, and the old gate was `isRunning = stats?.state === 'started'` with `TRANSITIONAL_STATES = {'archiving','stopping','starting'}` — so hosted already showed an enabled Stop over empty tabs in the same window. This PR closes it rather than opening it.

**`to_core_config()`** — confirmed, it deep-copies the MCP config per call. Fixed slightly wider than suggested: `CoreConfig(sandbox=self.sandbox, …)` passes the sandbox through by reference, so `_configured_provider` now reads `config.sandbox` directly and the deep copy is gone from the full path too, not just the offline one.

**`provider` across branches** — confirmed, and correctly flagged as currently harmless: both branches set `state` from the workspace row, and the endpoint only routes there when the row isn't `running`, so neither can structurally emit `running`. Set on all three branches anyway, with tests, so no consumer has to special-case a subset.

**Nullable `provider`** — fixed. `sandbox_id` was left alone: it is pre-existing `main` code and unrelated to this change.

Both new behaviours are mutation-tested rather than merely covered — reverting the sequence guard fails the new response-ordering test, reverting the offline clamp fails both new clamp tests, with checksum-verified restores. Full suites green: 7152 backend unit, 2534 frontend, `tsc --noEmit` and `ruff` clean.

Separately, `ff107fea` in this branch is unrelated to the review: `mcp 2.0.0` removed the `mcp.server.fastmcp` module scrapling imports, which is what broke `sandbox-docker`. It is pinned to `1.29.0` as a stopgap; the 2.x migration is a separate PR.

PR #344, opened 2026-08-04
fix(web): provider attribution on every web tool run, and the breaker bugs it exposed

Chen-zexi:
## Overview

| Category | Files | +LOC | −LOC |
|---|---:|---:|---:|
| Modified (excl. tests) | 13 | 244 | 60 |
| New (excl. tests) | 2 | 528 | 0 |
| Tests | 6 | 312 | 23 |
| **Total** | **21** | **1084** | **83** |
| **Net (excl. tests)** | **15** | | **+712** |

Two of those files are test tooling living outside `tests/` — `scripts/e2e/web_provider_pass.py` (486 lines, new) and a one-line docstring fix to its sibling `subagent_wire_pass.py` — so net *shipped* code is **+226** across 13 files.

### Breaker classification — the fix (net +96)
- `src/tools/web/providers/inhouse.py` (+52/−18) — crawler error map made total and stating `provider_fault` per outcome; an unmapped outcome warns instead of silently tripping the breaker
- `src/tools/web/breaker.py` (+50/−10) — breakers carry a name and their failure streak's causes; `on_open` fires once per transition into OPEN
- `src/tools/web/inhouse/safe_wrapper.py` (+21/−11) — the global infra breaker trips on `browser_closed` alone, and an unreachable target says so instead of blaming our infrastructure
- `src/tools/web/providers/firecrawl.py` (+13/−1) — an HTTP-200 `success: false` describes one URL's scrape, not an outage, so it no longer faults the provider

### Provider attribution — the feature (net +111)
- `src/observability/langsmith_run.py` (+42, new) — `stamp_run`, writes metadata/tags onto the enclosing LangSmith run; no-ops when tracing is off, never raises into the caller
- `src/tools/web/types.py` (+27/−1) — `FetchAttempt(provider, outcome)`; `WebError.native_kind` keeps the provider's own name for a failure; `providers_tried` kept as a derived property
- `src/tools/web/fetch.py` (+22/−5) — stamps at the single point every branch of the tool exits through
- `src/tools/decorators.py` (+12) — `run_metadata=` on `create_logged_tool`, for build-time choices the trace can't otherwise see
- `src/tools/web/crawl.py` (+10/−1), `src/tools/web/search.py` (+4/−1), `src/observability/__init__.py` (+2)

### Credit calibration (net 0, 5 values)
- `src/tools/manifest/web_providers.json` (+5/−5) — Firecrawl levels moved off an assumed 1:1 vendor-credit ratio onto measured consumption; Parallel's `fast` level corrected from 5 to 1

### Shared (net +19)
- `src/tools/web/router.py` (+25/−6) — names each provider's breaker and passes the failure reason (fix), and records each chain attempt's outcome (feature)

### Test tooling (+486 new, +1/−1)
- `scripts/e2e/web_provider_pass.py` — live agent pass; drives real flash and ptc turns, then asserts LangSmith attribution and breaker classification against a running stack
- `scripts/e2e/subagent_wire_pass.py` (+1/−1) — its usage example named a non-default port

**What this introduces:** provider attribution on every web tool run, so a fetch or search in a trace names what actually served it and why a chain fell through — plus the breaker-classification fix that attribution made visible, and a correction to two providers' credit values.

---

## Why

A fetch falls through a provider chain and a search engine comes from user prefs, both decided at runtime — but the trace kept no record of either. A run could not be attributed to a provider or filtered by one, which is what let two breaker bugs sit unnoticed.

### Bug 1 — target-side failures opened the shared provider breaker

`InhouseFetchAdapter` mapped every unlisted crawler outcome to `PROVIDER_ERROR`, and `PROVIDER_ERROR` counts against the provider's circuit breaker. So an unreachable host, a site that throttles, or a user cancelling a turn each counted as "the in-house provider is broken". Five of them disabled the in-house fallback for 60s, ratcheting toward a 900s cap.

The two routing axes are now orthogonal and both stated explicitly per outcome:

| | `retryable` | `provider_fault` |
|---|---|---|
| asks | try the next provider? | count against this provider's breaker? |
| true for | timeouts, anti-bot, empty, rate-limited | `queue_full`, `browser_closed` only |

The map is **total** — verified against every `error_type` `SafeCrawlerWrapper` can emit, and pinned by a test that fails if the crawler grows an outcome without a mapping. An unmapped outcome defaults to target-side and logs a warning, so the failure mode is a noisy log line rather than a silently disabled provider.

### Bug 2 — a few dead links fail-fast all crawling

Found by the E2E harness in this PR on its first run. The crawler's global infra breaker fails-fast *every* crawl while open, but it counted `dns_error`, `connection_refused` and the `infra_error` kind — all of which describe the target, not us. Five dead links in a research burst starved all crawling.

It now trips on `browser_closed` alone, which is genuinely ours and is the one thing its `on_open` browser-reset remedy actually fixes. A broken resolver and a domain that no longer exists are indistinguishable at that layer, and the two mistakes aren't equally cheap: a genuinely broken resolver fails every crawl anyway, so the breaker would only have saved the latency of trying. An unreachable host still opens its own per-host breaker.

### Bug 3 — Firecrawl's `success: false` faulted the provider

Same class as Bug 1, in the Firecrawl adapter: an HTTP 200 carrying `success: false` reports the outcome of scraping *that one URL* — a dead domain, a page no engine could render — but it mapped to `PROVIDER_ERROR` with `provider_fault` left at its default of `True`. A handful of dead links disabled Firecrawl outright. Verified against the live API: an unresolvable host returns exactly that shape, with `SCRAPE_DNS_RESOLUTION_ERROR`. A real Firecrawl outage arrives as a 5xx or a transport error and is still classified as a provider fault.

### Also

`on_open` fired once per failure rather than once per open. Calls already in flight when a breaker tripped kept recording failures, each re-running the recovery callback (a browser reset) and re-logging the same transition. It now fires once per transition *into* OPEN. A re-open out of HALF_OPEN still fires — the probe failed, so the remedy is owed another run — but failures recorded while already OPEN do not.

Breakers now carry a name and the causes of their current failure streak, so the one line a transition emits says *which* of a worker's breakers opened and why — `Circuit breaker [fetch:firecrawl] opening after 5 failures: 5x [rate_limited HTTP 429] Rate limit exceeded`. Repeats are collapsed (a breaker usually opens on five copies of one error), and routine failures stay silent.

## Credit calibration

Five values in `web_providers.json` changed. Each level should carry what a call actually consumes.

**Firecrawl** levels were set on the assumption that one Firecrawl credit maps to one of ours. Measuring the API directly gave the real ratios: a scrape consumes 1, a stealth scrape exactly 5, and map is flat per call however many links it returns — which confirmed the existing per-call unit for map was the right shape. The levels also carry headroom for failed scrapes, which consume a credit even though only successes are metered.

**Parallel's `fast` level** charged the same as `standard` and `deep` while sending `mode=turbo`, which the provider rates well below both. Beyond the overcharge, it left `fast` with no cost advantage over the levels above it — there was no reason to ever select it. Now corrected.

A sweep of the remaining providers found four smaller deviations (Tavily's research levels sit at their published band ceilings, Tavily fetch and Bocha are each slightly light, and Exa's fetch level is a flat value for a quantity that varies with the request shape). Those are left alone here — the Exa one needs a modelling decision rather than a new number — and are tracked as follow-up.

## What you get in a trace

`WebFetch` runs carry `fetch_provider`, `fetch_attempts`, `fetch_source`, `fetch_error` plus a filterable `fetch_provider:<p>` tag; `WebSearch` runs carry `search_engine` and `search_depth`. An attempt names the provider's own cause rather than the normalized bucket, so an unresolvable domain reads `inhouse:dns_error` and our own queue overflowing reads `inhouse:queue_full` — both previously `inhouse:provider_error`, which told a reader the provider was broken either way. A fallback reads `['firecrawl:budget_exceeded', 'inhouse:ok']` — the difference between "firecrawl was tried" and "firecrawl was out of budget", which is what makes a silent fallback legible after the fact. A degraded provider is now visible on the first call instead of staying silent until five failures open its breaker.

## Verification

Per repo convention, verified end-to-end against a live stack and live upstreams first; unit tests pin the settled contract.

- **Live agent pass.** Real flash and ptc turns against a running backend on the rebased tree. Breaker classification is green end to end: 7 unreachable hosts past a threshold of 5 left `inhouse` CLOSED at `failure_count=0`, the next healthy fetch was unpoisoned, and the server logged no shared-layer breaker transition while serving real traffic. Attribution is confirmed on the runs the pass produced — the flash turn's `WebFetch` recorded `fetch_provider=inhouse`, `fetch_source=live` and `fetch_attempts=['firecrawl:budget_exceeded', 'inhouse:ok']` with the `fetch_provider:inhouse` tag, both turns' `WebSearch` recorded `search_engine=tavily`/`search_depth=standard`, and the ptc turn's cache-served fetch correctly recorded `fetch_source=cache` and no provider.
- **Unit suite — 7649 passed**, 1 xfailed, 0 failed. Ruff clean; ESLint 0 errors.
- Bug 2 was caught by the harness before it was fixed, and the harness went green on the fix — the check that proves it is in the pass, not just in a unit test.
- Bug 3 was reproduced against the live Firecrawl API in both directions (faulting before the fix, not faulting after), which is what moved it from a documented residual to a shipped fix.

## Notes for review

- `providers_tried` survives as a derived property on `FetchResponse` — read only by tests now, but it's free and can't drift from `attempts`.
- The E2E pass fails rather than skips when it can't import `src`. A skipped breaker phase would report all-green while proving nothing, which is the exact failure this pass exists to catch.
- Firecrawl checks its rate limit *before* its credit balance, so an out-of-budget account presents as 429 under load. Worth knowing when reading `fetch_attempts` in production traces.
- The credit unit this manifest is denominated in isn't recorded anywhere in the file or its loader — worth a note there, since every future edit is otherwise a guess.

## Review round

- The attempt label collapsed a dead target and a broken provider into `provider_error` — the confusion Bug 1 is about, surviving in the telemetry this PR ships. `WebError` now carries the provider's native kind and the attempt prefers it. Found by probing a live `.invalid` fetch, not by either bot.
- The live pass graded every run in the LangSmith project's time window; it now correlates on the thread ids its own turns returned. On a machine running several stacks against one project it was judging other branches' runs.
- `stamp_run` de-duplicated an incoming tag against pre-existing tags only, so a repeat within one call landed twice. Unreachable from current call sites, fixed anyway.
- The E2E breaker phase discarded its fetch results, so its assertions could have passed without a target-side failure ever occurring; it now checks that first, and requires the exact `fetch_provider:<p>` tag rather than any tag with that prefix.
- A mutation check (suppressing the HALF_OPEN branch's `on_open`) left the whole suite green, so the callback behavior defended below had no coverage at all. Now pinned by a test that drives a failed recovery probe and fails under that same mutant.
- The E2E breaker phase returned an observation when the fetch chain resolved empty, so the pass could exit 0 with no breaker assertion evaluated — the same self-skip its own docstring forbids, and which its import-failure path already guards against. It now records a failed check.
- Declined: converting the sequential E2E driver to async (the convention governs the service; the one phase needing async already uses `asyncio.run`), and requiring a deterministic failed-primary scenario (it would fail the pass on any stack without a multi-provider chain configured — the assertion was tightened instead to require a served fall-through to end on `<provider>:ok`).
- `on_open` on a HALF_OPEN re-open is intentional, not a leak: the probe failed, so the browser reset is owed another run. Suppressing it would reset the browser once ever, across unlimited re-opens. The PR text above previously overstated this as "CLOSED→OPEN only" and has been corrected.

### DIFF
--- a/scripts/e2e/subagent_wire_pass.py
+++ b/scripts/e2e/subagent_wire_pass.py
@@ -25,7 +25,7 @@ Threads are kept for visual inspection (never deleted).
 
 Usage:
   uv run python scripts/e2e/subagent_wire_pass.py --user <user_id> \
-      [--workspace <ws_id>] [--base http://localhost:8020] [--model <model>] \
+      [--workspace <ws_id>] [--base http://localhost:8000] [--model <model>] \
       [--settle-timeout 600] [--turn-timeout 420]
 
 Auth: X-Service-Token from INTERNAL_SERVICE_TOKEN (repo .env or environment),
--- a/scripts/e2e/web_provider_pass.py
+++ b/scripts/e2e/web_provider_pass.py
@@ -1,0 +1,486 @@
+#!/usr/bin/env python3
+"""Web provider pass: agent turns → LangSmith attribution → breaker classification.
+
+Proves the two halves of the web-provider work against a live stack and live
+upstreams, driven by the real agent rather than by mocks:
+
+  1. Agent turns (flash and ptc, both by default): each turn is told to run a
+     web search and fetch a page, so WebSearch/WebFetch run inside a real
+     traced turn.
+  2. LangSmith attribution: the runs those turns produced — correlated by
+     thread id, since several stacks can share one project — must carry the
+     provider that actually served them — ``search_engine``/``search_depth``
+     on WebSearch, ``fetch_provider``/``fetch_attempts`` plus a
+     ``fetch_provider:<p>`` tag on WebFetch. A chain that fell through must
+     record WHY (``firecrawl:budget_exceeded`` → ``inhouse:ok``), not just
+     that it did.
+  3. Breaker classification (the fix): a URL that fails target-side, driven
+     past the failure threshold, must leave every provider breaker CLOSED with
+     a zero failure count, and must not poison the next healthy fetch. Runs
+     in-process because breaker state is process-local and no endpoint exposes
+     it; the network calls are real.
+  4. Server-side quiet check: the backend must not have logged a single
+     breaker transition while all of that happened — healthy failures are
+     silent by design, so any ``Circuit breaker [...]`` line is a regression.
+
+Phase 3 imports ``src`` and fails loudly if it cannot — it is the phase that
+proves the fix, so it must never quietly self-skip. Phase 4 needs docker and
+does skip with an observation when it is absent.
+
+Usage:
+  uv run python scripts/e2e/web_provider_pass.py --user local-dev-user \
+      [--workspace <ws_id>] [--base http://localhost:8000] \
+      [--agents flash,ptc] [--model <model>] [--turn-timeout 420]
+
+Auth: X-Service-Token from INTERNAL_SERVICE_TOKEN (repo .env or environment),
+user id via --user or E2E_USER_ID. Threads are kept for visual inspection.
+"""
+
+from __future__ import annotations
+
+import argparse
+import os
+import re
+import subprocess
+import sys
+import time
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+
+# A script's sys.path[0] is its own directory, never the cwd, so the repo root
+# has to be added explicitly — without it the breaker phase finds no ``src`` to
+# import and skips the one assertion that proves the fix.
+sys.path[:0] = [str(Path(__file__).resolve().parent), str(REPO_ROOT)]
+
+# The SSE/auth plumbing is identical for every wire pass — reuse it rather than
+# growing a second copy that can drift.
+from subagent_wire_pass import (  # noqa: E402
+    CHECKS,
+    Env,
+    check,
+    iter_frames,
+    observe,
+    read_env_file,
+)
+
+# A reserved TLD (RFC 2606) can never resolve, from any network, forever — the
+# only way to make "the target is unreachable" deterministic in a live test.
+UNREACHABLE = "https://nonexistent-host-{n}.invalid/"
+# Cache-busted per run: a cached URL never reaches the router, so it records no
+# provider at all — correct, but it would test nothing. The host has to tolerate
+# repeat automated traffic; example.com starts serving 403 after a few rounds,
+# which turns a provider-attribution assertion into a flake.
+HEALTHY = f"https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern?e2e={int(time.time())}"
+
+SEARCH_QUERY = "circuit breaker pattern in distributed systems"
+
+TURN_PROMPT = (
+    f"Do exactly two things, then stop. First, run ONE web search for "
+    f"'{SEARCH_QUERY}'. Second, fetch the page {HEALTHY} and quote its main "
+    f"heading. Reply in two sentences. Do not write any files or code."
+)
+
+ATTEMPT_RE = re.compile(r"^[a-z0-9_]+:[a-z0-9_]+$")
+
+
+# ---------------------------------------------------------------------------
+# Phase 1 — agent turns
+# ---------------------------------------------------------------------------
+
+
+def run_turn(env: Env, agent_mode: str, workspace_id: str, model: str | None) -> str | None:
+    """Post one turn and drain it. Returns the thread id (kept for inspection)."""
+    body: dict = {
+        "agent_mode": agent_mode,
+        "workspace_id": workspace_id,
+        "messages": [{"role": "user", "content": TURN_PROMPT}],
+    }
+    if model:
+        body["llm_model"] = model
+
+    outcome, thread_id, frames = None, None, 0
+    deadline = time.time() + env.turn_timeout
+    with env.client(read_timeout=env.turn_timeout) as c:
+        with c.stream("POST", "/api/v1/threads/messages", json=body) as resp:
+            if resp.status_code != 200:
+                resp.read()
+                check(f"{agent_mode}: turn accepted", False, f"HTTP {resp.status_code}: {resp.text[:200]}")
+                return None
+            for fr in iter_frames(resp):
+                frames += 1
+                thread_id = thread_id or str(fr.data.get("thread_id") or "") or None
+                if fr.event == "run_end":
+                    outcome = fr.data.get("outcome") or "completed"
+                    break
+                if fr.event == "error":
+                    outcome = "error"
+                    break
+                if time.time() > deadline:
+                    outcome = "timeout"
+                    break
+
+    check(
+        f"{agent_mode}: turn ran to completion",
+        outcome == "completed",
+        f"outcome={outcome} frames={frames} thread={thread_id}",
+    )
+    return thread_id
+
+
+# ---------------------------------------------------------------------------
+# Phase 2 — LangSmith attribution
+# ---------------------------------------------------------------------------
+
+
+def langsmith_checks(project: str, since: datetime, threads: list[str]) -> None:
+    try:
+        from langsmith import Client
+    except ImportError:
+        observe("langsmith SDK not installed — attribution checks skipped")
+        return
+    if not os.environ.get("LANGSMITH_API_KEY"):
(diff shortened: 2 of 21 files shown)


Chen-zexi:
Every finding was executed against the code rather than read, which changed two of the outcomes and surfaced a defect neither review caught. Pushed in `779477f4` and `255ebda3`.

## The one that mattered — found while verifying, not reported

Probing a live `.invalid` fetch returned `attempts=['inhouse:provider_error']`. `_attempt_outcome` built the label from `error.type.value`, but `WebErrorType.PROVIDER_ERROR` is a normalization bucket — `inhouse.py` maps `dns_error`, `connection_refused`, `network_error`, `crawl_error` and `infra_error` into it with `provider_fault=False`, alongside `queue_full`/`browser_closed` with `provider_fault=True`:

```
dns_error   (target is dead)  -> inhouse:provider_error
queue_full  (we are broken)   -> inhouse:provider_error
distinguishable in telemetry? False
```

The breaker classification was correct throughout — it keys off `provider_fault`. But the *label* is the deliverable here, and a trace of a dead domain read "the in-house provider is broken", which is exactly the confusion this PR exists to remove. `WebError` now carries `native_kind` and the attempt prefers it: `inhouse:dns_error` vs `inhouse:queue_full`.

## Accepted

- **Correlate LangSmith runs to this pass's turns.** Correct, and it was already producing false failures: several stacks here share one `LANGSMITH_PROJECT`, so the pass was grading other branches' runs. Now filters on the thread ids its own turns returned and requires each turn to yield both run types.
- **Unreachable URLs never checked for target-side failure.** Correct — the assertions could pass with no target-side failure having occurred. Results are now checked before the breaker state is.
- **Tag predicate accepts a mismatch.** Verified: `provider=inhouse` + tag `fetch_provider:firecrawl` passes today, fails under the exact predicate. Fixed.
- **`stamp_run` duplicate tags.** Verified: `['fetch_provider:a','fetch_provider:a']` → both appended; dedup across calls already worked, only within one list was broken. Unreachable from current call sites (all pass a single element), fixed anyway.
- **Nitpick — assert `attempts`/`providers_tried`.** Taken; it pins the compat shim for two lines.

## Declined

**`on_open` on HALF_OPEN→OPEN.** The premise is right, the conclusion isn't. Measured:

```
CLOSED->OPEN (2 failures)  : state=OPEN       callbacks=1
+2 failures while OPEN     : state=OPEN       callbacks=1   <- the fix working
after recovery timeout     : state=HALF_OPEN
HALF_OPEN probe fails      : state=OPEN       callbacks=2
```

The callback is `_trigger_browser_reset`. A failed half-open probe means the browser is still broken, so the remedy is owed another run. Suppressing it would reset the browser exactly once ever, across unlimited re-opens. The suppression that was needed — failures recorded while already OPEN — is in place, and the test is named for the real contract: once per transition *into* OPEN. The PR body said "CLOSED→OPEN edge only", which overstated it; that wording is corrected.

**Async-first E2E orchestration.** The convention governs the service, not a sequential test driver where phase ordering is the point. The one phase that genuinely needs async, `breaker_checks`, already uses `asyncio.run`.

**Mandatory failed-primary scenario.** Would fail the pass on any stack without a multi-provider chain configured; the `observe()` branch is deliberate. The fair half is taken — a served fall-through must now end on `<provider>:ok`.

**`deque(maxlen=failure_threshold)`.** Not fixable as suggested — `deque.maxlen` is read-only (`AttributeError: attribute 'maxlen' of 'collections.deque' objects is not writable`), so tracking a mutated threshold would mean rebuilding the deque on every change. Effect is one log line. Left alone.

## Known residual

The live label is now `inhouse:infra_error`, faithful to the crawler's own vocabulary — but that name is a misnomer: its only producer is the Tier-1 unreachable short-circuit, so it means *target unreachable*. Documented at `inhouse.py:42-45`. Renaming belongs in `scrapling_crawler.py`, out of scope here. `exa`, `parallel` and `tavily` collapse into `PROVIDER_ERROR` the same way and would benefit from `native_kind`; this PR doesn't otherwise touch them.

7652 unit tests pass, ruff clean.

PR #350, opened 2026-08-14
refactor(quota): relay denials from the quota service, and fail closed on no verdict

Chen-zexi:
The credit gate held rules it cannot get right: its own BYOK negative-balance threshold, cached in Redis for 60s, and a frontend that re-authored a denial message per limit type from a table that had fallen behind the service. This reduces the gate to what a client should do.

1. **Relay denials instead of re-authoring them.** The 429 path forwards the service's message, limit type and counts untouched. `byok` becomes a flag we report, not a branch we take. Deletes the local debt rule, its verdict cache, and the frontend copy table (which covered 5 of 9 limit types, and told everyone their quota "resets at midnight UTC").
2. **Fail closed when there is no verdict.** An unanswered credit check raises 503 `service_unavailable` instead of admitting the turn. Only this gate changes posture; the six other callers still fail open, unchanged. It retries once on a transport failure, an unreadable 200 or a 5xx, and not on a 4xx. An automation denied by that 503 does not count a strike toward auto-disable.

Two denials stay ours: the concurrency gate, and the unavailable-service guard.

> **Deploy ordering: the paired quota-service change must ship first.** Details below.

## Overview

|  | Files | + | − | Net |
|---|---:|---:|---:|---:|
| Production | 9 | +236 | −187 | **+49** |
| Tests | 8 | +531 | −280 | +251 |
| **Total** | **17** | **+767** | **−467** | **+300** |

Production is net +49 across 9 modified files: mostly deletion, paid back in comments recording what was deleted and why.

- **Backend** — `dependencies/usage_limits.py` (the gate), `app/threads/{create,messaging}.py` (pass `byok` through), `services/automation_executor.py` (503 is our outage, not a strike).
- **Frontend** — `utils/rateLimitError.ts` (relay), `components/ui/error-banner.tsx`, `api/client.ts` (lift `detail.message` onto the axios error at any status), two call sites.
- **Tests** — one new (`transport.errors.test.ts`, locking the link between a gate's HTTP response and what the user reads, off a 503 body captured live). One deleted: `MarketPanelError.test.tsx` asserted against a copy of the production logic pasted into the test file, so it could not fail on a change to the real code.

## Test plan

- Backend gate suites → 98 passed; `ruff` clean
- Web suites for the touched files → 40 passed; `pnpm typecheck` clean
- Each new test verified by reverting its fix and confirming it fails

Not yet verified end to end: a BYOK denial rendered in the browser against the *new* quota service, which is not deployed yet. That is the case the deploy ordering is about; it should be checked in staging once the quota service lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quota and credit-limit handling during service outages with consistent validation, retry guidance, and clearer 503 errors.
  * Workflow executions affected by quota-service outages are marked failed without triggering automatic disablement.
  * Server-provided error messages are now preserved in the interface.

* **User Experience**
  * Error banners support multiple links, including external links that open in new tabs.
  * Rate-limit and quota errors now provide more relevant guidance and navigation options, including usage and plan links when applicable.


Chen-zexi:
Blocking finding is correct and now fixed in `0f85462a`.

`resp.json()` sat outside the transport `try`, so an unreadable 200 escaped `_call_validate_for_user` entirely. It is worse than described in one respect: it also escapes the six fail-open callers, which have no handler at all, so a proxy's HTML error page turned every quota check in the service into a 500. And `origin/main` had `resp.json()` inside the one `try/except Exception`, so this refactor genuinely traded fail-open for a crash rather than exposing a pre-existing hole.

Fixed by rejoining the no-verdict path: an unreadable 200 sets `reason` and `continue`s, so strict callers retry once then raise `QuotaServiceUnavailable` (→ 503 + `Retry-After`), and everyone else gets `None` as before.

Three tests, each verified to fail without the fix:
- `test_an_unreadable_200_is_no_verdict_and_still_retried` — raises `QuotaServiceUnavailable`, `post` awaited twice
- `test_an_unreadable_200_still_fails_open_for_everyone_else` — returns `None`
- `TestCreditGateFailsClosed::test_an_unreadable_200_blocks_the_turn_as_a_503` — the user-visible contract: 503, `service_unavailable`, `Retry-After: 15`. Without it the turn is still refused, but as a 500 with no `Retry-After`, and the automation executor counts it as a strike because it only exempts 503.

Backend gate suites: 98 passed, ruff clean.

---

**`/account` fallback — not taking it.** Three reasons:

1. Not in this diff. Those lines are pre-existing; the hunks at `useChatMessages.ts:1745` and `useMarketChat.ts:590` only rename `link` → `links`.
2. It is the documented default, not a stopgap. `web/AGENTS.md` specifies `VITE_PLATFORM_URL` default `/account`, because in the standard deploy the account portal is same-origin at `/account/*`. Passing `undefined` there would drop the CTA on the deployment shape that is supposed to have one.
3. The OSS case it describes is unreachable. With `HOST_MODE=oss` the platform gate is a no-op, so no quota denial carrying a portal CTA is ever produced; `burst_limit` is ours and already returns no links. `AccountMenu.tsx:32` handles the one place an OSS fork could see such a link, by hiding it outright.

PR #363, opened 2026-08-22
feat(plugins): Agent Plugins standard support and the reworked Plugins page

Chen-zexi:
## Overview

Adds support for the [Agent Plugins](https://agent-plugins.org) open standard (v1.0.0) and rebuilds the Plugins page around it. A package declares MCP servers (`mcp.json`) and agent skills (`skills/`); installing it, from a zip or a public git URL, fans those components into the per-user tables that already existed, each row stamped with its plugin's identity.

**Totals**

| | Files | +lines | −lines |
|---|---|---:|---:|
| Total | 176 | 18,086 | 1,183 |
| Net (excl. tests) | 117 | 14,371 | 1,102 |

**By module** (net lines, largest first)

| Module | Files | Net | Notes |
|---|---|---|---|
| `web/src/pages/Plugins/` | 39 new, 6 mod | +5,794 −707 | Plugins tab and install wizard, origin decks, detail overlays, list controls, bulk selection, provenance badges |
| `src/server/services/plugins/` | 23 new | +4,754 | The install pipeline: fetch, archive, discover, validate, grants, fan out, update, reconcile, export, probe, plus two vendored schemas |
| `src/server/app/` | 1 new, 6 mod | +732 −143 | Plugins router, user-tier vault blueprints endpoint, per-server tools endpoint, provenance projection on the catalog routers |
| `src/server/database/` | 1 new, 2 mod | +813 −100 | `plugins.py` (entity, caps, row lock) plus the join predicate and provenance columns on the server and skill queries |
| `web/src/pages/ChatAgent/` | 2 new, 13 mod | +693 −53 | Shared MCP primitives the Plugins page reuses; the status pill made the single exhaustive status table |
| `web/src/locales/` | 2 mod | +344 −10 | The `plugins` namespace in both catalogs |
| `web/src/hooks/` | 1 new, 4 mod | +325 −35 | Query hooks and the fan-out invalidator; `useDialogA11y` grew dialog stacking and backdrop dismissal |
| `src/server/services/` (other) | 1 new, 7 mod | +314 −19 | Catalog-edit seam, shared OAuth grant-revoke helper, generalized pinned stream client |
| `src/server/models/` | 1 new, 1 mod | +257 −15 | Install report, component result, diagnostic wire shapes |
| `migrations/` | 2 new | +240 | `028_agent_plugins` (the `user_plugins` table and the provenance FKs) and `029_plugin_owned_secrets` |
| `web/` (other) | 2 mod | +59 −2 | Query keys and one token |
| `src/ptc_agent/` | 1 mod | +41 −18 | The shared credential scanner, so install-time extraction and export-time scrub agree field for field |
| Tests | 48 new, 11 mod | +3,715 −81 | Conformance suite (one fixture per failure-ladder rung, each asserting the *scope* of the failure), router and database tests, frontend badge, dialog and list tests |

**What it introduces**

- **Install from a zip or a public git URL.** Forge URLs normalize to an HTTPS archive endpoint, streamed under a pin and a byte cap; extraction carries member, ratio, and containment guards.
- **Marketplace and vendor layouts.** Discovery reads `marketplace.json` indexes and the `.claude-plugin` / `.codex-plugin` / `.cursor-plugin` manifests, adapting dialects to canonical form in memory. A source holding several plugins returns a chooser rather than an error.
- **A plugin is one switch.** Turning it off withdraws every component from delivery at once; editing one takes it out of the plugin's hands permanently; updating re-fetches and reconciles.
- **Declared credentials resolve, and the report says which ones were already yours.** A plugin may reference any name your vault holds; when it reaches for one you set up for something else, the install report names the credential and the endpoint it is going to.
- **Export.** A spec-compliant zip regenerated from the stored manifests and per-skill archives, with embedded credentials blanked.
- **The page is reworked.** Collapsible decks by origin, a monogram tile and URL-addressable detail overlay per row, and one shared search, filter, select mode and Add menu across the three tabs.

## Notable design decisions

**A plugin is a wrapper, not a fourth config tier.** The MCP resolver already merges three tiers and that algebra is the hard part of the system; a fourth would multiply its cases in the hot path of every turn. Instead the resolver is never told plugins exist: components carry `plugin_id`, and plugin-level disable is one join predicate, `(plugin_id IS NULL OR p.enabled)`, on the two queries that deliver them. Since that predicate is the whole enforcement, `adopt_server` (which moves a user-tier row down into a workspace, below the predicate) now refuses plugin-owned rows the same way it refuses OAuth-connected ones.

**Editing a component detaches it.** Customize clears `plugin_id` in place, rather than letting a later update merge over a change the user made deliberately. It costs a warning on the edit response and a `"detached"` line in the next update report.

**A plugin may use the user's own credentials, and has to say so.** A manifest can name any credential and the reference resolves. The vault is the user's, they installed the plugin deliberately, and the same package on a local coding agent would resolve it without asking anyone. Refusing bought a narrower blast radius than the sandbox already has, and paid for it in the failure that actually happens: a server that installs looking healthy and then 401s on its first call, over a name as ordinary as `GITHUB_TOKEN`.

What the split buys instead is disclosure. Bind validation stays separate from materialization, and after materialization the entry configs are the complete statement of what will be referenced, so a name the plugin introduced passes quietly while one the vault already held is reported with the destination it is going to. The portable `mcp.json` can write `${vault:NAME}` into a header or a url with no extension block at all, so judging the configs rather than the manifest is what makes both doors answer to the one rule.

**Neither the stdio command nor the shape of a reference is filtered.** A command is launched with an argv list and no shell, inside the same sandbox where the agent already runs commands on the user's behalf, so an allowlist bounded nothing that was not already bounded; what it decided was which published servers could be installed at all, and `docker`, `deno`, `bun` and a plain binary path are ordinary ways to ship one. Likewise a reference may sit inside a larger value: `Authorization: Bearer ${vault:TOKEN}` is the shape nearly every server documents, and the sandbox has always substituted references in place rather than replacing the field.

**The provenance foreign keys are `ON DELETE RESTRICT`, on `(user_id, plugin_id)`.** A cascade would delete component rows while skipping the purges their delete helpers do unconditionally (workspace tombstones squatting a `UNIQUE` slot, discovery schema rows a same-name recreate would resurrect, archive objects). `SET NULL` is the detach semantic, a different operation. So uninstall walks the existing helpers first and deletes the plugin row last, and the constraints name the user alongside the plugin because every writer already scopes by user, which is exactly what would make a query that forgets invisible. Vault secrets are the one exception, and take `SET NULL` deliberately: uninstalling unclaims a credential rather than destroying it, so a reinstall never silently drops a value the user typed.

**Untrusted input is pinned or vendored, never followed.** `git` is not in the backend image and a subprocess clone re-resolves DNS after any pre-connect check, so sources are fetched as HTTPS archive snapshots with every hop re-pinned through `pin_public_url`. The canonical JSON schemas are vendored and selected against an empty `referencing.Registry()`, so a remote `$ref` in a hostile manifest resolves to nothing; an opt-in integration test diffs the vendored bytes against the canonical URLs.

**Install is deliberately not one transaction.** A crash leaves partial components and update reconciles them. The alternative holds a connection for the length of a network fetch, across archive extraction, object storage writes and N component creations.

One deliberate deviation from the spec, a legal per-entry skip reported with a `spec_ref` in the diagnostics rather than hidden: `${PLUGIN_ROOT}` / `${PLUGIN_DATA}` entries are held back because the sandbox does not materialize the plugin tree yet, so the path they name would not exist. `cwd` is accepted and ignored.

## What the ecosystem actually publishes

The last stretch of this branch was QA against real sources rather than fixtures, installing from `openai/plugins`, `wshobson/agents`, `BuilderIO/skills` and others in all four dialects. Six things only real packages surface:

- **A root manifest can be a cover page.** `wshobson/agents` ships a root `.cursor-plugin/plugin.json` carrying a name, a version and no components, beside a marketplace index listing 91 real plugins. Root-wins installed the cover page: a 201, a plugin slot spent, and nothing to run. Nothing in the tree tells that apart from a genuine root plugin, so the traversal now answers both questions and the caller decides, offering the chooser when a person is waiting and reconciling the same plugin when an update is not asking anyone.
- **A package built for something else installs, and says so.** An editor extension or a hosted app packaged as a plugin has no better candidate to fall through to, so it installs with a warning naming what it did contain.
- **Naming a subdirectory has to reach nested plugins.** Root-wins is suspended when the caller names a path, which is what makes the 91 offers above resolvable instead of a chooser that refuses every path it just listed.
- **The skills cap was sized for hand-uploads.** One plugin can ship 30 or more skills, so 50 per account silently truncated the second one installed. Raised to 200 and counted over the right set.
- **An update that changed nothing said "updated".** Update rebuilds every config from scratch, so an untouched entry still reached the write looking new, spending two workspace-wide MCP version bumps to deliver an identical row. It now compares against the committed row and reports `unchanged`, matching what the skill arm always did.
- **Warnings spoke as the validator.** Every dialect puts something in `plugin.json` the canonical schema does not model, so "Additional properties are not allowed ('displayName', 'skills' were unexpected)" was the warning nearly every real install showed. It now names the keys and says they are kept verbatim, which is what the spec asks for.

## How it works

`src/server/services/plugins/` is a pipeline, split so everything a fatal problem could refuse is decided before the first write. `fetch.py` and `archive.py` get the tree in safely; `discover.py` walks it for canonical and vendor layouts; `manifest.py` / `mcp.py` / `extension.py` implement the spec's failure ladder literally, so a `plugin.json` violation writes nothing, a document defect drops only the MCP component, and an entry defect skips only that entry. `package.py` is the boundary: it returns a `ValidatedPackage` and nothing downstream re-parses the archive. `grants.py` keeps disclosing a reference separate from materializing it, and the fan-outs reuse the existing MCP import machinery so caps, reserved names, duplicate detection and vault extraction all apply unchanged.

The seam worth reading closely is `src/server/services/mcp_catalog.py`. Fork-on-edit used to live inside the database writer, so the update path had to pass a flag asking the writer not to detach the row it was updating, which was the source of a 500 on every in-place update. The detach now lives in `apply_catalog_edit`, the writer is dumb, and the plugin path calls it directly.

Install and export were the two paths whose cost scaled with a package's skill count, one thread hop and one storage round trip each, bounded only by the per-user skill cap. Export now carries the archive locator on the enumeration; install splits into a concurrent phase that decides each plan and stores its archive, and a serial one that writes the rows.

On the frontend, `web/src/pages/Plugins/` keeps rows in `components/` and behaviour in `hooks/`, so the three list surfaces share one selection machine, one bulk runner and one deck. `utils/provenance.ts` and `components/PluginBadges.tsx` are the single source for whether a row is plugin-owned and whether its plugin is off, so a call site cannot forget the suppression guard, and the same predicate gates every delete affordance.

## Commits

Eight, sliced by area rather than by the order they were written:

1. `feat(plugins): the plugin entity and component ownership`
2. `feat(plugins): validate a package against the Agent Plugins schemas`
3. `feat(plugins): install, update, and uninstall a validated package`
4. `feat(plugins): the plugins API`
5. `test(plugins): conformance fixtures and pipeline pins`
6. `feat(web): provenance primitives and the plugins data layer`
7. `feat(web): the Plugins page`
8. `test(web): pin the Plugins page surfaces`

## Test plan

- [x] `uv run pytest tests/unit/`: **8,929 passed**, 1 xpassed, 0 failures
- [x] `cd web && pnpm test`: **285 files, 3,087 tests passed**; `pnpm typecheck` clean
- [x] `make lint`: 0 errors
- [x] Live E2E against a running stack: installed from real public sources in all four dialects (`openai/plugins`, `wshobson/agents`, `BuilderIO/skills`, `shadcn-ui/ui`, `gitroomhq/postiz-agent` and others). **49 assertions green on the final tree**: export round trip 16, empty-package handling 13, no-op update reporting 15, plugin-disable projection 5. The install, duplicate-install, chooser, fork-on-edit, update, uninstall, credential-binding and export flows were also driven through the browser UI
- [x] Every commit's tree resolves its own imports: every `src.*` and every `@/`/relative TypeScript specifier checked against that commit's tree, all eight clean

Not covered: visual QA on the reworked page beyond the flows exercised above, and two code-verified-only changes outside it (the workspace MCP panel's "Connectors" label, and a vault blueprint badge stepping 12px to 10px to match provenance badges elsewhere).

---
<a href="https://app.blacksmith.sh/ginlix-ai/codesmith/LangAlpha/pr/363"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789973700&installation_model_id=432753&pr_number=363&repository=ginlix-ai%2FLangAlpha&return_to=https%3A%2F%2Fgithub.com%2Fginlix-ai%2FLangAlpha%2Fpull%2F363&signature=ef1f1db809770d98a95073c9730d7185324d654ae26cbcfa4f764b54c5d39afb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Chen-zexi:
Verified all 11 at `dda7ff21`. Seven reproduced as defects and are fixed; three are rebutted; one was a product call and is now implemented.

**Fixed**

1. **`McpServers.tsx:353` selectDisabled** — confirmed. `allWorkspaceServers` is read at :151 and rendered as `workspaceSections` at :443, and `useMcpBulkActions` takes `workspaceServers` and dispatches `tier: 'workspace'`, so the feature was fully built and unreachable. Fixed as suggested.

2. **`probe.py` pin outside the deadline** — confirmed, and the bound really is unenforced: `resolve_public_ips` calls `loop.getaddrinfo` with no timeout of its own. Moved the pin inside `asyncio.timeout`. Patched the resolver to hang and re-ran: the probe returns in 1.00s against `PROBE_TIMEOUT_S=1` instead of stalling its semaphore slot. Also gave `TimeoutError` its own arm, because `str()` on one is empty and the broad handler reported the failure the caps exist to produce as no reason at all.

3. **`vault_invalidation.py` N+1** — confirmed. Hoisted the row read into `after_secrets_changed` and passed it down; a fetch failure hands the loop `None` so the per-name fallback bump keeps its failure domain. Measured with a counting fake tier: 6 names now issue 1 server-list query, down from 6, with all 6 bumps intact.

6. **`grants.py` `secret_names_in` misses `url`** — confirmed, and worth more than defense in depth: `_resolve_http` in the sandbox client resolves `${vault:NAME}` in the URL exactly as in headers, so `validate_remote_url` really is the only thing in the way. Added `url` to both the scan and `strip_refs` (stripping the ref text, as args do), with the coupling written down. Verified: a config with refs in url and headers now reports both, and stripping one leaves the other.

5. **`SCHEMAS_BY_ID`** — confirmed dead, one occurrence in the repo. Deleted.

8. **`plugins.ts` export error path** — confirmed. Re-parse the error blob in place rather than teaching the shared formatter about blobs, so a refused export shows the backend's detail instead of `Request failed with status code N`.

9. **`PluginReportStep.tsx` LANDED** — half confirmed. `detached` under "Skipped" is correct (it was skipped, and carries the reason). `error` was not: it fell through the same negation into neutral text. Errored components now render in the existing errors group, no new copy.

**Product call, implemented (10)** — `spec_ref` was unrendered, confirmed: one occurrence in `web/src`, the type declaration. It is now a citation link on every diagnostic that carries one, via a shared `DiagnosticLine` used by the report and the step errors. Seven of the nine diagnostics the fixture suite produces carry a link, so it is live rather than decorative.

**Rebutted**

4. **`manifest.py` bare `PluginFatal`** — the comparison does not hold. `mcp.py`'s version check is not a fatal at all; it returns a `_component_error` because a document defect drops only the MCP component. `StepError` renders `error` as the primary line and diagnostics as sub-lines, so the message is what the user reads and the array being empty costs nothing visible. Structural only, and all three bare raises in that file are consistent with each other.

7. **`set_plugin_enabled` duplicating `_bump_user_versions`** — `_bump_user_versions` is private to `mcp_servers.py`. Importing another module's underscore-prefixed helper to save four lines of SQL trades a real boundary for a cosmetic one.

11. **Memoization** — a settings page with a handful of rows per deck; the same pattern predates the Plugins page, as noted. Not spending a round on it.

**Verified**: `ruff check src/` clean · `pytest tests/unit/` 8906 passed · `tsc --noEmit` clean · `vitest` 3081 passed (285 files) · targeted repros for findings 2, 3, 6 and 10 as described above. Visual QA on 1 and 9 was not run: this worktree's frontend is in platform mode and login-gated, so those two are verified by reading, with the type and unit gates green.

Chen-zexi:
Confirmed and fixed in `ce9d561f`, and the finding was bigger than the symlink path.

**Reproduction.** Same 99.8 KiB zip, one member carrying a 100 MiB DEFLATE stream, only the mode bits differ:

| member | verdict | time | peak |
|---|---|---|---|
| regular file | refused `compression_ratio` | 0.00s | 0.1 MiB |
| `S_IFLNK` | accepted | 0.07s | 209.5 MiB |

**Two more holes of the same class, found while verifying yours.**

The premise that the ratio guard holds for regular files does not survive a dishonest header. `ZipFile.read(info)` does not respect the declared `file_size` at all: it inflates the whole stream and only then compares the CRC. Patching the uncompressed-size field to 20 in both the local header and the central directory made the ratio check pass (20 against a cap of 20,384,600) and `zf.read(info)` still inflated 209 MiB before raising `BadZipFile`. So the regular-file path was defeatable by a lying header, and the refusal arrived after the damage rather than instead of it.

Tar is affected too, just not through `linkname`. `tarfile` inflates a GNU long-name payload while *parsing the header*, before the member reaches any check in our loop. A 398 KiB `.tar.gz` declaring a 400 MiB member name allocated 2,000 MiB and escaped as an unhandled `OSError`, which the router turns into a 500 for what is a malformed upload.

**The fix**, since all three share one root cause: guards computed from attacker-written header metadata, followed by an unbounded read.

Every cap is now applied twice. Once from the header, so an honest bomb is still refused without inflating a byte, and again on the bytes that actually came out, which is the half a lying header cannot walk past. `_inflate()` reads through `zf.open(info).read(limit + 1)`, which bounds the inflation itself, with the limit set to whichever of the remaining byte budget and the ratio allowance is nearer. A link target gets a path-sized cap of its own (`MAX_LINK_TARGET_BYTES = 4096`) rather than the file-sized one, since a link target is a path. The tar stream is unwrapped behind a `_CappedReader` instead of by `tarfile`, so every read `tarfile` makes is bounded, header-extension reads included.

I did not take the suggested "cap `info.file_size` to a few KB", on its own: that is one more guard reading the number the archive wrote, and it is exactly what the lying-header case defeats.

| case | before | after |
|---|---|---|
| symlink-marked bomb | accepted, 209.5 MiB | refused `link_target_too_long`, 0.1 MiB |
| lying-header bomb | refused after 209.5 MiB | refused after 39.1 MiB |
| tar long-name bomb | `OSError` (500), 2,000 MiB | refused `too_large`, 0.3 MiB honest / 400 MiB worst case |

**Verification.** Three pins in `TestArchiveHardening`, each ablated against the pre-fix tree to confirm it fails there: the symlink bomb installed, the lying-header bomb peaked at 45 MiB against a 16 MiB payload, and the tar case raised `OSError`. Full backend units 8932 passed, `ruff check src/` clean. Live E2E through the router, 11/11, including a real `openai/plugins` forge tarball install (the only production path that runs the tar branch) and a vendor-alias symlink package, which is the reason links are resolved rather than refused in the first place.

Issue #364, opened 2026-08-23
DeepSeek weekends have billed off-peak all day since 2026-08-23; resolve_schedule still charges peak


xyzs996:
`resolve_schedule` picks the window from `at.astimezone(timezone.utc).hour` alone. DeepSeek changed the rule at 00:00 Beijing on 2026-08-23 (`2026-08-22T16:00:00Z`): the whole of Saturday and Sunday now bills off-peak, peak hours included, with the weekday windows unchanged.

So since yesterday, a `deepseek-v4-pro` call completing on a weekend inside `[1,4)` or `[6,10)` UTC is priced at $1.32/$3.96 when it was charged $0.66/$1.98 — double, in the fail-high direction the rest of this function deliberately chooses, which is why nothing here will surface it.

Two things that make this less trivial than a `weekday()` check:

**The weekend is bounded in Beijing time**, not UTC — it runs 16:00Z Friday to 16:00Z Sunday. The two calendars disagree only over 16:00–24:00Z, and both of DeepSeek's windows sit clear of that. So `at.weekday()` on the UTC instant returns the correct tier for all 168 hours of the week today, and starts mispricing the day a window moves past 16:00Z. Anything testing it needs instants at the 16:00Z edges, where the hour is off-peak either way.

**It is not retroactive.** Weekend peak hours before `2026-08-22T16:00:00Z` were charged at peak. Since `resolve_schedule` takes `at` and is used on recorded completion stamps, a rule without an effective-from would reprice every weekend already on record.

Given the shape here I would guess the card grows something like

```json
"off_peak_days": {"days": [5, 6], "utc_offset_minutes": 480, "from": "2026-08-22T16:00:00Z"}
```

validated in `_usable_windows`'s neighbourhood and failing high the same way, rather than a DeepSeek constant in `pricing_utils`. But the manifest shape is yours — happy to send the PR either way if you say which.

Sources: <https://api-docs.deepseek.com/quick_start/pricing>. Vectors for both axes as plain JSON: <https://github.com/xyzs996/deepseek-peak-offpeak-vectors>

PR #365, opened 2026-08-23
fix(pricing): count DeepSeek peak days on the vendor clock, not UTC

xyzs996:
## Summary

DeepSeek's off-peak discount is published as a **weekday** rule on Beijing time, but the two DeepSeek cards in `providers.json` only carry the hour windows. A request that lands inside a peak window on a Beijing weekend is billed at full rate today, even though the vendor charges off-peak for it.

This adds an optional day restriction to the rate card and reads the day on the vendor's own clock. Windows stay declared and read in UTC — only the *day* is shifted, which is the only part the vendor states in local terms.

Worth flagging plainly: with the windows as published today, both readings of the calendar price **identically at all 168 hours of the week**, because the two peak windows sit clear of the hours where a UTC day and a Beijing day disagree. So no test written against the real windows can tell the two apart. That is exactly why the tests below use a synthetic `[[15, 20]]` card — it's the only window shape that makes the split visible, and it pins the rule before a future window edit walks into it silently.

## Changes

- `pricing_utils.py`: new `peak_days` (ISO 1–7) and `peak_days_utc_offset` on the rate card; `resolve_schedule` charges peak only when the instant is both in a window and on a peak day.
- Both keys added to `_SCHEDULE_KEYS`, so they're stripped from the priced card like the other scheduling keys and never reach a consumer as a rate.
- `providers.json`: `deepseek-v4-pro` and `deepseek-v4-flash` get `"peak_days": [1,2,3,4,5]`, `"peak_days_utc_offset": 8`.
- Validation fails **high**, not low: an unreadable day list drops the restriction entirely rather than dropping bad entries one by one. Dropping entries would narrow the peak days and hand out a discount nobody earned; dropping the restriction only ever charges the current, published rate. Same for a bad offset — it falls back to reading the day off UTC.
- `type(d) is int` rather than `isinstance`, because JSON `true` is an `int` in Python and would otherwise install Monday as a peak day.

## Test Plan

- `tests/unit/llms/test_peak_hour_pricing.py`: 8 new tests in `TestPeakDays`, including `test_the_day_is_read_on_the_vendors_clock_not_utc` — 2026-01-02 15:59Z is peak and 16:00Z is off-peak (UTC Friday becomes Beijing Saturday), 2026-01-04 16:00Z is peak again (UTC Sunday becomes Beijing Monday). Plus: a card with no `peak_days` stays unrestricted, days without an offset are counted in UTC, seven shapes of unreadable day list (`[]`, `"1-5"`, `[0]`, `[8]`, `[True]`, `[1,"2"]`, `5`) all charge peak, and neither key reaches the rate card.
- `tests/unit/llms/test_manifest_integrity.py`: shape check on the two new keys, and it rejects an offset declared without a day list — on its own it shifts a clock nothing reads, so it looks like a restriction that isn't there.
- Mutation-checked: removing the day condition from the peak branch reddens exactly the four day-axis tests (4 failed, 27 passed); restoring it gives 31 passed.

One caveat on how that was run: the full suite needs `langchain_openai` and `pytest_asyncio`, which I couldn't install here, so I ran the scheduling tests against `pricing_utils.py` loaded by path in isolation — 31 passed (your 23 plus my 8). The manifest-integrity check was validated by running its new assertion logic against the shipped `providers.json` directly. Both are worth a normal CI run before you take my word for them.

---
<a href="https://app.blacksmith.sh/ginlix-ai/codesmith/LangAlpha/pr/365"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790049805&installation_model_id=432753&pr_number=365&repository=ginlix-ai%2FLangAlpha&return_to=https%3A%2F%2Fgithub.com%2Fginlix-ai%2FLangAlpha%2Fpull%2F365&signature=87ed95b5b5541140de11f6eb9b1cc79658293d11366f45d296e11979b2e3aa6c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added weekday-aware peak pricing for DeepSeek V4 Pro and DeepSeek V4 Flash.
  * Peak pricing applies Monday through Friday based on vendor-local time (UTC+8).
  * Pricing schedules now support configurable weekday restrictions and timezone offsets.

* **Bug Fixes**
  * Improved handling of invalid pricing schedule settings, preventing incorrect peak-rate application.

### DIFF
--- a/src/llms/manifest/providers.json
+++ b/src/llms/manifest/providers.json
@@ -386,6 +386,14 @@
               10
             ]
           ],
+          "peak_days": [
+            1,
+            2,
+            3,
+            4,
+            5
+          ],
+          "peak_days_utc_offset": 8,
           "off_peak": {
             "input": 0.66,
             "cached_input": 0.022,
@@ -415,6 +423,14 @@
               10
             ]
           ],
+          "peak_days": [
+            1,
+            2,
+            3,
+            4,
+            5
+          ],
+          "peak_days_utc_offset": 8,
           "off_peak": {
             "input": 0.22,
             "cached_input": 0.007,
--- a/src/llms/pricing_utils.py
+++ b/src/llms/pricing_utils.py
@@ -2,7 +2,7 @@
 Pricing calculation utilities for LLM token usage with support for both flat and tiered pricing.
 """
 
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import Dict, Any, Optional, List, Tuple
 import logging
 import re
@@ -12,7 +12,13 @@ logger = logging.getLogger(__name__)
 # Keys describing *when* a card applies rather than what it charges. Stripped
 # from the card handed to the engine so what gets priced, logged and snapshotted
 # is only the rates in force.
-_SCHEDULE_KEYS = ("peak_utc", "off_peak", "schedule_anchor")
+_SCHEDULE_KEYS = (
+    "peak_utc",
+    "off_peak",
+    "schedule_anchor",
+    "peak_days",
+    "peak_days_utc_offset",
+)
 
 # An unstamped record's model name is the vendor's string; bound it before it
 # reaches a log line.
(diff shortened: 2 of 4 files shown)


Chen-zexi (approved):

Chen-zexi:
Thanks! @xyzs996

Chen-zexi:
this pr close #364

PR #369, opened 2026-08-25
fix(prefs): stop an empty model catalog from wiping saved model preferences

Chen-zexi:
## Overview

| | Files | Insertions | Deletions |
|---|---:|---:|---:|
| Production | 2 | 70 | 16 |
| Tests | 3 | 227 | 1 |
| **Total** | **5** | **297** | **17** |

**By module**

| Module | File | What changed |
|---|---|---|
| `src/server/services/llm` | `availability.py` | the stale-preference scrub returns early when the model manifest is empty, and the removed-model error stops claiming a preference was cleared when none was |
| `web/src/pages/Settings/panels` | `ModelTab.tsx` | `saveModelPrefs` and the sibling auto-cleanup effect gate on the raw catalog rather than on the access-filtered `validModelNames`; the effect also gains the arrays it cleans as dependencies |

**What this introduces:** two guards against one bug class. An empty model catalog now reads as "not loaded" rather than "zero models exist", so neither the backend scrub nor the frontend save deletes a user's saved model preferences while the manifest is unavailable.

## The bug

Both sites treat an empty catalog as ground truth, so if the manifest fails to load every saved model name looks stale.

- **Backend.** `_cleanup_stale_model_preferences` deletes each unrecognised name. Those deletes go to a merge-upsert where `None` means "remove this key", so the whole preference set goes with no copy kept.
- **Frontend.** `saveModelPrefs` validates against `validModelNames`. Empty means `preferred_model`, `preferred_flash_model`, `compaction_model`, `fetch_model`, `starred_models` and `fallback_models` are all nulled in a single write. It is debounced and fires on any Model tab edit, including changing the search provider, and also flushes on unmount.

Neither has an undo. The sibling effect in `ModelTab.tsx` already carried a guard of the same shape, which is what made the omission visible — though that guard turned out to have the hole described below.

`validModelNames` is the wrong thing to ask, in two independent ways, both surfaced in review.

It is **not a loaded signal.** `useAllModels` folds the user's own custom models into it, so a user with one custom model has a non-empty set while the server catalog is still unavailable, and every saved built-in reads as stale.

It is **not an existence signal either.** `buildVisibleModels` branches on the platform access response and only the truthy side fails open: with `platform === null` it runs the configured-provider filter, which admits nothing for a user with no BYOK keys. Since `usePlatformModels` sets `retry: false`, a single failed `/api/auth/models` leaves that set empty for the whole session against a fully loaded catalog. Keying cleanup on access is wrong on its own terms too — a lapsed tier or a removed key would read as a deleted model.

A third problem sat in the same effect, found in review. It listed neither array it cleans as a dependency, so it only ever ran before `loadModelTabData`'s await resolved and never saw the stored values. That is invisible for starred models, which render filtered, but the fallback chips render straight from state — `FallbackModelsPicker` maps `selected` with no catalog filter — so a model that no longer existed stayed on screen until the models query happened to refetch. The deps are now exhaustive and the eslint-disable is gone.

So the two questions are now asked separately: `catalogLoaded`, from the raw `/api/v1/models` payload, for whether the catalog arrived at all; `knownModelNames`, from the pre-filter catalog plus custom models, for whether a name exists. This is what the backend already does — `mc.get_model_config(name)` reads the process-global manifest, never a per-user filtered view. Display filtering still uses `validModelNames`: a model the user cannot reach should not render, it should just not be deleted.

## Verification

Reproduced end to end against a running stack. Emptying `src/llms/manifest/models.json` makes `/api/v1/models` answer HTTP 200 with zero providers and zero models, which is exactly the condition both call sites were missing: a successful response carrying nothing.

With the manifest emptied, changing the Web Search Provider in Settings sent, on `main`, `preferred_model: null`, `preferred_flash_model: null`, `starred_models: null` and `fallback_models: []`, and the database confirmed the wipe. On this branch the same action sent every name intact.

Both review findings were covered the same way. One custom model with an empty catalog keeps every built-in; a loaded catalog with empty access data keeps every built-in. Each was confirmed by reverting the gate and watching that test alone fail, which is the only way to know it tests anything.

- Backend: scoped tests pass, `ruff check` clean.
- Frontend: 27 Settings tests pass, eslint and `tsc --noEmit` clean.

The frontend tests are a set on purpose: nothing is dropped while the catalog is empty; a genuinely stale name is still dropped once it has loaded; a custom model alone does not count as a loaded catalog; an empty access set does not count as an empty catalog; and a stale fallback chip does clear once the catalog is in. Without the second, this fix would be indistinguishable from deleting the cleanup.

## Follow-up, not in scope

The cleanup effect and `saveModelPrefs` now filter the same two arrays through the same predicate. One of them should own that, rather than both.

## Risk

Both paths now skip cleanup when the catalog is unavailable, so a genuinely stale name survives until the next save that happens with a loaded catalog. That is the intended trade: a delayed cleanup is recoverable, a deleted preference set is not.

---
<a href="https://app.blacksmith.sh/ginlix-ai/codesmith/LangAlpha/pr/369"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790282767&installation_model_id=432753&pr_number=369&repository=ginlix-ai%2FLangAlpha&return_to=https%3A%2F%2Fgithub.com%2Fginlix-ai%2FLangAlpha%2Fpull%2F369&signature=4d09523094615f443b531954410a74210f12254cb1305e34f1e55f8c50a21cb2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved saved model preferences while the model catalog is unavailable or still loading.
  - Prevented valid model references from being removed when model data is incomplete.
  - Continued removing preferences for models no longer present after the catalog loads.
  - Improved guidance when a selected model is unavailable, without incorrectly indicating that a preference was deleted.

- **Tests**
  - Added coverage for catalog loading states, stale model cleanup, and fallback selections.

### DIFF
--- a/src/server/services/llm/availability.py
+++ b/src/server/services/llm/availability.py
@@ -67,6 +67,16 @@ async def _cleanup_stale_model_preferences(user_id: str) -> list[tuple[str, str]
     pref = await get_model_preference(user_id)
 
     mc = LLMFactory.get_model_config()
+    # An empty manifest is a load failure, not a catalog of zero models. Every
+    # name would then read as stale and this function would delete the user's
+    # entire model preference set — irreversibly, since the deletes go straight
+    # to the merge-upsert with no copy kept.
+    if not mc.llm_config:
+        logger.warning(
+            f"[CHAT] Skipping stale-pref scrub for user={user_id}: model manifest is empty"
+        )
+        return []
+
     custom_models = {cm.get("name") for cm in (pref.get("custom_models") or [])}
     custom_providers = {cp.get("name") for cp in (pref.get("custom_providers") or [])}
 
@@ -125,14 +135,21 @@ def _raise_model_removed(
 
     other = sorted({name for _, name in removed if name != model_name})
     extra = f" Also cleared: {', '.join(other)}." if other else ""
+    # Only claim a clear when one happened. The caller passes an empty list for
+    # a model named in the request rather than saved, and the scrub also returns
+    # empty when it declines to run, so asserting the clear unconditionally
+    # tells the user their preference is gone while it is still stored.
+    cleared = (
+        "Your saved preference has been cleared — open Settings to pick a current model."
+        if removed
+        else "Open Settings to pick a current model."
+    )
 
     raise HTTPException(
         status_code=400,
         detail={
             "message": (
-                f"Model '{model_name}' is no longer available. "
-                "Your saved preference has been cleared — open Settings to pick a current model."
-                + extra
+                f"Model '{model_name}' is no longer available. " + cleared + extra
             ),
             "type": "model_removed",
             "link": {"url": "/settings?tab=model", "label": "Open Settings"},
--- a/tests/unit/server/services/llm/test_stale_pref_scrub_guard.py
+++ b/tests/unit/server/services/llm/test_stale_pref_scrub_guard.py
@@ -1,0 +1,63 @@
+"""The stale-model-preference scrubber must not act on an empty catalog.
+
+Regression: ``resolvable()`` asks the manifest whether a model name still
+exists. When the manifest fails to load, every name answers "no" and the
+scrubber deletes the user's whole model preference set through a merge-upsert
+that keeps no copy. There is no undo, so the guard is the fix.
+"""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from src.server.services.llm.availability import _cleanup_stale_model_preferences
+
+STORED_PREFS = {
+    "preferred_model": "some-model",
+    "preferred_flash_model": "some-flash",
+    "fetch_model": "some-fetch",
+    "compaction_model": "some-compaction",
+    "fallback_models": ["fallback-a", "fallback-b"],
+    "profiles": {"some-model": {"reasoning_effort": "high"}},
+}
+
+
+def _catalog(names):
+    mc = MagicMock()
+    mc.llm_config = {n: {"model_id": n} for n in names}
+    mc.get_model_config.side_effect = lambda n: mc.llm_config.get(n)
+    return mc
+
+
+async def _run(catalog_names):
+    upsert = AsyncMock()
+    with (
+        patch("src.llms.llm.LLM.get_model_config", return_value=_catalog(catalog_names)),
+        patch(
+            "src.server.services.llm.config.get_model_preference",
+            AsyncMock(return_value=dict(STORED_PREFS)),
+        ),
+        patch("src.server.database.user.invalidate_user_prefs_cache", AsyncMock()),
+        patch("src.server.database.user.upsert_user_preferences", upsert),
+        patch("ptc_agent.agent.graph.invalidate_user_profile_cache", AsyncMock()),
+    ):
+        removed = await _cleanup_stale_model_preferences("user-1")
+    return removed, upsert
+
+
+@pytest.mark.asyncio
+async def test_empty_manifest_scrubs_nothing_and_writes_nothing():
+    """The destructive case. An empty catalog means "not loaded", not "all gone"."""
+    removed, upsert = await _run([])
+    assert removed == []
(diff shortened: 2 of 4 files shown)


Chen-zexi:
Both findings valid, both fixed. Branch resliced so each fix sits in the commit it corrects; head is now `09e8c2b6`.

**P1 — gate on the catalog, not the merged set.** Correct, and it was my bug: `useFilteredModels.buildVisibleModels` folds custom models into the map (creating the provider group if absent) before `validModelNames` is built, so one custom model makes the set non-empty against an empty server catalog and every saved built-in reads as stale. `ModelTab.tsx` now derives `catalogLoaded` from the raw `/api/v1/models` payload and gates on that.

The same hole was in the sibling auto-cleanup effect, which the finding did not name — it had the original `validModelNames.size === 0` guard and would have wiped the same set on mount. Fixed alongside.

Pinned by a third test: one custom model, empty catalog, every built-in preserved. Reverting the gate to `validModelNames.size > 0` makes it fail (1 failed / 18 passed), restoring it passes 19.

**P2 — do not claim a clear that did not happen.** Correct. `_raise_model_removed` asserted the preference had been cleared unconditionally, and there are two paths where `removed` is empty: the new early return, and a model named in the request rather than saved, which was already wrong before this branch. The CTA is now conditional on `removed` being non-empty; the "open Settings to pick a current model" half stays either way.

25 Settings tests, 2 backend guard tests, `ruff`, eslint and `tsc --noEmit` all clean.

Chen-zexi:
Valid, and the exposure is larger than a cold-load race. Fixed in `2e0b3cc7`.

`buildVisibleModels` (`useFilteredModels.ts:332-339`) branches on `platform` and only the truthy side fails open. With `platform === null` it runs `filterModelsByAccess`, which admits nothing when `configuredSet` is empty, so a platform-mode user with no BYOK keys gets an empty `validModelNames` against a fully loaded catalog. `filterByPlatformTier`'s own `if (!platform) return providerMap` is unreachable from there, which is why the fail-open docstring on `usePlatformModels` does not describe the pipeline.

`usePlatformModels` also sets `retry: false`, so this is not only the window before the response lands. One failed `/api/auth/models` leaves `platform` null for the session and every subsequent Model tab edit wipes that user's preferences.

Took the second of your two suggestions rather than the first. Gating on readiness of all access inputs fixes the timing but leaves the worse half standing: keyed on access, a lapsed tier or a deleted BYOK key reads as "this model no longer exists" and nulls the preference for a model that never went anywhere. The two questions are now asked separately — `catalogLoaded` from the raw `/api/v1/models` payload for whether the catalog arrived, `knownModelNames` from `rawModels` (pre-filter catalog plus custom models) for whether a name exists. That also removes the dependency on platform-query readiness, so no new loading-state plumbing.

This matches the backend, which already validates against the process-global manifest via `mc.get_model_config(name)` and never against a per-user filtered view. Checked for the same hole there; it does not exist.

Pinned by a fourth test: platform mode, catalog loaded, `validModelNames` empty, every saved name preserved. Reverting `known()` to `validModelNames` fails that test alone (1 failed / 19 passed); restoring passes 20. Display filtering still uses `validModelNames` — an unreachable model should not render, it should just not be deleted.

26 Settings tests, `ruff`, eslint and `tsc --noEmit` clean.

PR #392, opened 2026-09-04
feat(llms): add GPT-6 Astra and bump the OpenAI SDK stack

Chen-zexi:
## Summary

Adds GPT-6 Astra to the model catalog on both OpenAI routes, and bumps the OpenAI SDK stack to a version that speaks the API Astra ships against.

Astra is offered as `gpt-6-astra` on the API-key route and `gpt-6-astra-oauth` on the ChatGPT Codex route, both at the 1.05M context window with a 128K output ceiling. Its reasoning ladder is `low, medium, high, xhigh, max` with a `medium` default, one rung shorter than the GPT-5.6 family: the endpoint rejects `none` and `minimal` for this model, so offering them would be four buttons where two of them 400.

## Overview

| | Files | + | − | Net |
|---|---|---|---|---|
| Manifests | 2 modified | 106 | 0 | +106 |
| Dependencies | 1 modified | 5 | 2 | +3 |
| Tests | 1 modified | 15 | 10 | +5 |
| Lockfile (generated) | 1 regenerated | 61 | 63 | −2 |
| **Total** | **5 modified, 0 new** | **187** | **75** | **+112** |

Net excluding tests and the lockfile: **+109**.

**By module**

- `src/llms/manifest/models.json` (modified, +68) — two new entries. `gpt-6-astra` on `openai` carries `prompt_cache_options: {mode: implicit}` like its GPT-5.6 siblings; `gpt-6-astra-oauth` on `codex-oauth` carries `include: [reasoning.encrypted_content]` with `store: false` instead. Both declare the five-rung ladder and `prompt_guidance: lean`.
- `src/llms/manifest/providers.json` (modified, +38) — one pricing entry under `openai`, using the same short/long-context threshold shape as the GPT-5.6 entries. `$10.00` input, `$1.00` cached input, `$12.50` cache write and `$50.00` output per 1M, with input and cache rates doubling and output at 1.5x past 272K input tokens. The `codex-oauth` twin carries no pricing list and inherits these through the parent provider.
- `tests/unit/llms/test_model_pricing_inheritance.py` (modified, +15/−10) — the threshold-billed OpenAI ids move into a shared `_TIERED_OPENAI_IDS` tuple that `gpt-6-astra` joins, so the existing structural assertions on tiered input, input-dependent output and codex-oauth inheritance now cover it. The assertions stay value-free, so a reprice does not touch them.
- `pyproject.toml` (modified, +5/−2) — `langchain-openai>=1.6.0`, `langchain-core>=1.6.0`, and `openai>=3.8.0` declared directly, since `src/llms/vendor/langchain_zai` and `libs/ptc-cli` import it rather than reaching it through langchain-openai.
- `uv.lock` (regenerated, +61/−63) — `openai` 2.45.0 → 3.8.0, `langchain-openai` 1.4.1 → 1.6.0, `langchain-core` 1.5.3 → 1.6.1, `jiter` 0.15.0 → 0.16.0.

**What this introduces:** GPT-6 Astra becomes selectable in the composer with a working effort control, on an API key or a connected ChatGPT account, and its usage is costed correctly including the long-context threshold and the cache read and write rates.

## Verification

Against the live endpoint, not the model card:

- **Reasoning ladder, on both routes.** Probing each level returns `Unsupported value: 'none' is not supported with the 'gpt-6-astra' model. Supported values are: 'low', 'medium', 'high', 'xhigh', and 'max'.` for both `none` and `minimal`, and accepts the other five. Identical on `api.openai.com` and on the Codex backend, so the rejection is the model's rather than one route's policy. The manifest declares exactly those five.
- **Every rung runs on the API-key route.** All five completed against `api.openai.com`, each echoing its own effort back, with reasoning tokens rising across the ladder.
- **Effort reaches the wire.** Building the entry through `ModelSpec.from_manifest` and running `apply_reasoning_effort` for each rung emits five distinct payloads, and each one was accepted by the endpoint.
- **Image input.** A real turn with an `input_image` part was answered and billed image tokens, so the declared modality is not aspirational.
- **Reasoning preservation on the OAuth route.** `encrypted_content` rides back on `response.output_item.done` identically to `gpt-5.6-sol`, which is what the entry's `include` plus `store: false` shape depends on.
- **Cost path.** A 100K input / 40K cache read / 50K cache write payload through `extract_token_usage` into `calculate_total_cost` charges reads at 0.1x and writes at 1.25x and subtracts both from plain input, matching the rates in the entry.
- **After the SDK bump.** Full unit suite green (9702 passed), `ruff check src/` clean, and a real turn through `LLM(...)` on the new entry.

## Judgment calls worth a second opinion

- **`speed: 2` and `intelligence: 5`** are editorial ratings for the model picker. `speed` was measured rather than taken from the vendor card, which rates Astra "Fast": across three streamed runs at `medium` effort on the same prompt, Astra averaged about 37 output tokens/sec against 62 for GPT-5.6 Sol, 60 for Terra and 70 for Luna, making it the slowest of the four. On the existing scale (Luna 4, Terra 4, Sol 3) that lands at 2, alongside `claude-opus-5` and `deepseek-v4-pro`. `intelligence: 5` matches its GPT-5.6 siblings.
- **`pdf` in `input_modalities`** mirrors the GPT-5.6 family rather than a probe. Image input was verified; PDF was not.
- **The 272K threshold** is read as applying to the full request, so both tiers carry their own `cached_input` and `cache_5m`, matching how the GPT-5.6 entries were authored.
- **`openai` 3.x is a major bump.** Its one breaking change is the HTTPX2 migration, which affects callers that pass the OpenAI client a custom httpx client or transport. We pass none, and our own httpx usage is unrelated to the SDK. The visible consequence is that `opentelemetry-instrumentation-httpx` no longer sees the SDK's own requests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,8 +17,11 @@ dependencies = [
     "httpx[http2]>=0.28.1",
     # LangChain core packages (v1)
     "langchain>=1.0.0",
-    "langchain-core>=1.4.7",
-    "langchain-openai>=1.3.5",
+    "langchain-core>=1.6.0",
+    "langchain-openai>=1.6.0",
+    # Declared directly because src/llms/vendor/langchain_zai and libs/ptc-cli
+    # import it; the 3.x floor is the HTTPX2 client migration.
+    "openai>=3.8.0",
     "langchain-anthropic>=1.0.0",
     "langchain_google_genai>=2.0.0",
     "langchain-deepseek>=0.1.4",
--- a/src/llms/manifest/models.json
+++ b/src/llms/manifest/models.json
@@ -134,6 +134,40 @@
       "write": "parameters.reasoning.effort"
     }
   },
+  "gpt-6-astra": {
+    "prompt_guidance": "lean",
+    "model_id": "gpt-6-astra",
+    "provider": "openai",
+    "display_name": "GPT-6 Astra",
+    "visible": true,
+    "speed": 2,
+    "intelligence": 5,
+    "context": 1050000,
+    "input_modalities": [
+      "text",
+      "image",
+      "pdf"
+    ],
+    "parameters": {
+      "reasoning": {
+        "summary": "auto"
+      },
+      "prompt_cache_options": {
+        "mode": "implicit"
+      }
+    },
+    "reasoning": {
+      "efforts": [
+        "low",
+        "medium",
+        "high",
+        "xhigh",
+        "max"
+      ],
+      "default": "medium",
+      "write": "parameters.reasoning.effort"
+    }
+  },
   "claude-fable-5": {
     "prompt_guidance": "lean",
     "model_id": "claude-fable-5",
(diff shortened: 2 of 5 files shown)


Chen-zexi:
TPS is measured live, PR body updated. `speed: 2`  is the correct value

PR #395, opened 2026-09-05
fix(web): settle a hidden tab's streaming animations on the first frame back

Chen-zexi:
## Summary

Returning to a streaming chat after a tab switch no longer shows the catch-up glitch. Before, the first frames back were a burst of live rows collapsing, a scroll sweep across the whole gap, and reasoning text that re-typed itself. After, the first painted frame is the settled layout, in place, with one cursor.

### What a user sees

| Before | After |
|---|---|
| Switch tabs mid-reasoning, come back: every row that folded while away collapses in front of you, one after another. | The rows are already folded in the first frame back. |
| The view slides from where it was to the new bottom over a second or so. | The view is at the bottom in the first frame back. |
| The reasoning text flip-flops by a few hundred characters for several seconds, as if two cursors were typing the same row. | One cursor, monotonic text. |

### Root causes

Three, all in the web client, all verified over raw CDP on real Chrome against the live stack (a Playwright page never reports hidden; the CDP script drives a second real tab).

- **framer queues animations in a hidden tab.** Its frameloop is rAF-driven, so an animation created while hidden sits at its first keyframe and plays in full on the first frame back. A streaming transcript queues dozens.
- **framer's `stop()` can complete synchronously.** `JSAnimation.stop()` ticks the wall clock before tearing down; a chain older than its duration (routine when a hidden tab's effects run once a second) finishes inside `stop()` and its `onComplete` starts a sibling typewriter chain the hook no longer tracks.
- **The follow scroll trailed the transcript** with a smooth `scrollTo` that a hidden tab never animates, so the first follow after return swept the whole gap.

### How

- `lib/hiddenTabMotion.ts` sets `MotionGlobalConfig.skipAnimations = document.hidden` on every visibility change, so animations created while hidden are instant. In-flight animations at hide time need nothing: their first tick back reads the wall clock and finishes them in that frame.
- `useAnimatedText` stamps each chain with a generation (`chainRef`); a superseded chain's `onUpdate` and `onComplete` return without touching state.
- `useChatScroll` jumps instantly while hidden and re-pins on `visibilitychange`, once synchronously and once on the next frame, after the instant animations have applied.

### Measured

Same live turn, tab hidden for 12 s mid-reasoning, sampled per DOM mutation after return:

| | main | this branch |
|---|---|---|
| non-monotonic transcript mutations after return | 53 | 0 |
| queued row collapses played on return | dozens | 0 |

The regression test `animated-text.stop.test.ts` drives the hook under a hidden document with a clock-based framer mock and asserts one live chain; it fails on main with two.

## Overview

| | Files | Added | Removed |
|---|---|---|---|
| Modified | 3 | 49 | 12 |
| New | 2 | 88 | 0 |
| Total | 5 | 137 | 12 |
| Net excluding tests | 4 | 74 | 12 |

- **Web, animation**: `lib/hiddenTabMotion.ts` (new, the skipAnimations toggle), `main.tsx` (side-effect import), `components/ui/animated-text.ts` (chain generation guard).
- **Web, scroll**: `pages/ChatAgent/components/chatView/useChatScroll.ts` (instant follow while hidden, first-frame re-pin on return).
- **Tests**: `components/ui/__tests__/animated-text.stop.test.ts` (new).

## Risk

- `skipAnimations` is global to framer. Any animation created while the tab is hidden, anywhere in the app, is instant. Nothing visible is painting then, so nothing is lost.
- The re-pin on return only fires when the reader was near the bottom and no pin owns the position.

---
<a href="https://app.blacksmith.sh/ginlix-ai/codesmith/LangAlpha/pr/395"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791179002&installation_model_id=432753&pr_number=395&repository=ginlix-ai%2FLangAlpha&return_to=https%3A%2F%2Fgithub.com%2Fginlix-ai%2FLangAlpha%2Fpull%2F395&signature=5aa3d55d55c27253368ee771c8ef9e8accaaed2585097cc12b376e8773eb160b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Animations now complete instantly when the app is opened or updated in a hidden browser tab.
  - Chat transcripts automatically catch up to the latest content when returning to the tab, while preserving the position of users who have scrolled away.
  - Streaming chat content uses instant scrolling in hidden tabs and smooth scrolling when visible.

- **Bug Fixes**
  - Prevented animated text from duplicating or restarting unexpectedly during rapid updates and tab visibility changes.

### DIFF
--- a/web/src/components/ui/__tests__/animated-text.stop.test.ts
+++ b/web/src/components/ui/__tests__/animated-text.stop.test.ts
@@ -1,0 +1,63 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+
+// framer's stop() runs one last tick at the wall clock before tearing down, so
+// a chain older than its duration finishes inside stop() and fires onComplete
+// synchronously, and a stop() on an already finished animation fires it
+// again (JSAnimation.tick treats state "finished" as done). Routine in a
+// hidden tab, where the effect re-runs once a second. This mock reproduces
+// exactly that, with a clock the test advances by hand.
+type Opts = { duration: number; onUpdate: (v: number) => void; onComplete: () => void };
+let clock = 0;
+const live = new Set<object>();
+vi.mock('framer-motion', () => ({
+  animate: (from: number, to: number, opts: Opts) => {
+    const createdAt = clock;
+    let finished = false;
+    const controls = {
+      stop: () => {
+        live.delete(controls);
+        const elapsed = (clock - createdAt) / 1000;
+        if (finished || elapsed >= opts.duration) {
+          finished = true;
+          opts.onUpdate(to);
+          opts.onComplete();
+        } else {
+          opts.onUpdate(from + (to - from) * (elapsed / opts.duration));
+        }
+      },
+    };
+    live.add(controls);
+    return controls;
+  },
+}));
+
+import { useAnimatedText } from '../animated-text';
+
+const words = (n: number) => Array.from({ length: n }, (_, i) => `w${i}`).join(' ') + ' ';
+
+describe('useAnimatedText chain ownership', () => {
+  afterEach(() => {
+    Object.defineProperty(document, 'hidden', { configurable: true, value: false });
+  });
+
+  it('a chain finished by its own stop() never starts a sibling', () => {
+    const { result, rerender } = renderHook(({ text }) => useAnimatedText(text, { enabled: true }), {
+      initialProps: { text: 'seed ' },
+    });
+    // Hidden, seconds between updates (Chrome wakes a hidden tab once a second
+    // at best, once a minute under intensive throttling), more than a snap's
+    // worth of text in each: the cleanup stop finishes the chain, then the
+    // backlog snap stops it again after the target has moved on. That second
+    // stop used to spawn a sibling chain aimed at the old target.
+    Object.defineProperty(document, 'hidden', { configurable: true, value: true });
+    let text = 'seed ';
+    for (let i = 0; i < 6; i++) {
+      clock += 3000;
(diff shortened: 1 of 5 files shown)


Chen-zexi:
Both non-blocking items addressed or answered.

**rAF in `handleVisibility`: fixed** (amended into the commit, now `6c2580c`). The deferred frame is tracked in `visibilityRafRef`, replaced if a second visibility flip lands before it fires, and cancelled in the effect cleanup, matching the `entryRestoreRafRef` pattern.

**`skipAnimations` blast radius: intended, and the PR's Risk section states it is global.** The check you asked for: the app has 61 framer-motion consumers and exactly one completion callback outside the typewriter, `AutomationInlineForm.tsx:204`, which flips `overflow` to visible once its height animation completes. Under the flag that callback fires on the first frame back instead of after a played-out animation, which is the same end state sooner. No toast, timer or dismissal is tied to an animation completing. The alternative, an animation created while hidden, is one that stalls at its first keyframe until the tab returns and then plays in full, so instant completion is the safer default for every surface, not just the transcript.

